PR #65, opened 2020-08-27
useMapInteraction

nerik:
Screenshot 2020-08-27 at 16 32 35

Adds a useMapInteraction hook whose responsibility is to pick a clicked or hovered feature and deliver information about these features in a standardized way.
These features can then be passed to Layer Composer's global config. This PR contains an example of how a generator can pick that up to generate a distinct visual style for a selected or highlighted feature.
Additionally, this PR adds a transformation which injects `interactiveLayerIds` to the style's metadata section. This looks at `metadata.interactive` for each layer and collect layer's ids. `style.metadata.interactiveLayerIds` can then be passed directly to react-map-gl

Example of use: 
https://github.com/GlobalFishingWatch/react-map-gl/pull/26

TODO 
- how to define interaction priority on layers for click/hover?
- use the hook to define which cursor should be used
- use the hook to define popup data in a standardized way


packages/layer-composer/src/generators/heatmap/heatmap-animated.ts
```diff
@@ -95,36 +97,53 @@ class HeatmapAnimatedGenerator {
...
-      const source = {
-        id: timeChunk.id,
-        type: 'temporalgrid',
-        tiles: [decodeURI(url.toString())],
-        maxzoom: config.maxZoom,
-      }
-      return source
-    })
+    const sources = flatten(
+      timeChunks.map((timeChunk: TimeChunk) => {
```
j8seangel, line 101: instead of using two methods and depends on lodash, why not using [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) instead?

Also has a filter method builtin, because returning an empty array will remove the element from the final output

packages/react-hooks/src/use-map-interaction/use-map-interaction.ts
```diff
@@ -0,0 +1,146 @@
...
+                  if (extendedFeature.value === 0) {
+                    extendedFeature = null
+                  }
+                } else {
+                  extendedFeature = null
+                }
+            }
+            return extendedFeature
+          })
+          .filter((f: ExtendedFeature) => f !== null)
```
j8seangel, line 86: example of a simpler filter version with flatMap could apply here

j8seangel (approved):
Amazing work! it will simplify a lot the process and it is a good starting point a couple of comments to talk about them later on:

- I like a lot the hover state styles are contained in the [generator](https://github.com/GlobalFishingWatch/react-map-gl/pull/26/files#diff-5d2b71ad4d4dfc563ad4bc7d771d2788R84), but we will need to thing a way to customize 
- We will need to have the current value in `SourcesWithHoverState` as for example to show the current value in the legend

packages/layer-composer/src/generators/heatmap/heatmap-animated.ts
nerik: Didn't know that, thanks!!

applications/temporalgrid-demo/src/App.js
```diff
@@ -60,48 +105,66 @@ export default function App() {
...
+          colorRamps = ['bivariate']
+        }
+          
         generators.push({
           id: 'heatmap-animated',
           type: Generators.Type.HeatmapAnimated,
-          tilesets: activeTilesets.map(t => t.tileset),
-          filters: activeTilesets.map(t => t.filter),
+          sublayers: heatmapSublayers,
+          sublayersCombinationMode: combinationMode,
```
j8seangel, line 160: is there a combinationMode not applying to sublayers? it could be redundant

applications/temporalgrid-demo/src/App.js
```diff
@@ -113,50 +176,67 @@ export default function App() {
...
-  // TODO switch between debounced/immediate/throttled time when using animated
-  const { style } = useLayerComposer(layers, (animated) ? time: debouncedTime)
+  const [mapRef, setMapRef] = useState(null)
+
+  const [clickedEvent, setClickedEvent] = useState(null)
+  const clickCallback = useCallback((event) => {  
+    // probably dispatch a redux action here or whatever
+    setClickedEvent(event)
+    // TODO
+    // if (feature.popupCallbackURL) {
```
j8seangel, line 196: nice TODO

applications/temporalgrid-demo/src/popup.js
```diff
@@ -0,0 +1,41 @@
+import { Popup } from '@globalfishingwatch/react-map-gl';
+import React from 'react';
+
+function PopupWrapper ({ tooltipEvent, closeButton, closeOnClick, className, onClose, children }) {
```
j8seangel, line 4: I think its worth having this component either react-map-gl custom fork or ui-components to be able to consume hoverTooltipEvent as simple as possible.

j8seangel (commented):
Let's talk about some pending thinks and align in the next steps but in general lgtme

applications/temporalgrid-demo/src/App.js
```diff
@@ -113,50 +176,67 @@ export default function App() {
...
+  const [hoveredEvent, setHoveredEvent] = useState(null)
+  const hoverCallback = useCallback((event) => {
+    setHoveredEvent(event)
+  }, [])
 
-  const mapRef = useRef(null)
-  if (mapRef && mapRef.current) {
-    mapRef.current.getMap().showTileBoundaries = debug
-    mapRef.current.getMap().on('idle', () =>  {
+  const { onMapClick, onMapHover } = useMapInteraction(clickCallback, hoverCallback, mapRef)
```
j8seangel, line 209: I think we could split by callback in the same way we did for the useTooltip for the different event, so in case we only want hover we don't have to send clickCallback as undefined

applications/temporalgrid-demo/src/App.js
j8seangel: cool story bro

PR #72, opened 2020-09-08
4wings multiple layers

j8seangel:
Use stats by `config.id` to allow having different 4wings layers at the same time
Also include param called `steps` to be able to customise steps from the generator instead of the stats (normally saved in the dataview)

### DIFF
--- a/packages/layer-composer/src/generators/heatmap/config.ts
+++ b/packages/layer-composer/src/generators/heatmap/config.ts
@@ -26,17 +26,33 @@ export const HEATMAP_GEOM_TYPES_GL_TYPES: HeatmapGeomGL = {
   [HEATMAP_GEOM_TYPES.EXTRUDED]: 'fill-extrusion',
 }
 
+const opacitySteps = [0, 0.2, 0.4, 0.6, 0.8, 1]
+const hex2Rgb = (hex: string) => {
+  const cleanHex = hex.replace('#', '')
+  const color = {
+    r: parseInt(cleanHex.slice(0, 2), 16),
+    g: parseInt(cleanHex.slice(2, 4), 16),
+    b: parseInt(cleanHex.slice(4, 6), 16),
+  }
+  return `${color.r}, ${color.g}, ${color.b}`
+}
+
+const getColorRampByOpacitySteps = (finalColor: string) => {
+  const color = finalColor.includes('#') ? hex2Rgb(finalColor) : finalColor
+  return opacitySteps.map((opacity) => `rgba(${color}, ${opacity})`)
+}
+
 export const HEATMAP_COLOR_RAMPS: Record<ColorRampsIds, string[]> = {
   fishing: [DEFAULT_BASEMAP_COLOR, '#3B9088', '#EEFF00', '#ffffff'],
   presence: [DEFAULT_BASEMAP_COLOR, '#163F89', '#0F6F97', '#07BBAE', '#00FFC3', '#FFFFFF'],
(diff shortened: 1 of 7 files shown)


packages/layer-composer/src/generators/heatmap/config.ts
```diff
@@ -26,17 +26,33 @@ export const HEATMAP_GEOM_TYPES_GL_TYPES: HeatmapGeomGL = {
...
-  teal: [DEFAULT_BASEMAP_COLOR, '#004F6B', '#007B7F', '#00A793', '#00D3A7', '#00FFBC'],
-  magenta: [DEFAULT_BASEMAP_COLOR, '#33306E', '#663D86', '#994A9E', '#994A9E', '#FF64CE'],
-  lilac: [DEFAULT_BASEMAP_COLOR, '#1F3D78', '#3E579A', '#5E70BB', '#7D8ADD', '#9CA4FF'],
-  salmon: [DEFAULT_BASEMAP_COLOR, '#333F64', '#665B72', '#665B72', '#CC928D', '#FFAE9B'],
-  sky: [DEFAULT_BASEMAP_COLOR, '#004C78', '#00749A', '#009DBB', '#00C5DD', '#00EEFF'],
-  red: [DEFAULT_BASEMAP_COLOR, '#333156', '#663E56', '#994C54', '#CC5A54', '#FF6854'],
-  yellow: [DEFAULT_BASEMAP_COLOR, '#334B45', '#667334', '#999A22', '#CCC211', '#FFEA00'],
-  green: [DEFAULT_BASEMAP_COLOR, '#214F57', '#427B58', '#64A757', '#85D358', '#A6FF59'],
-  orange: [DEFAULT_BASEMAP_COLOR, '#333E48', '#665939', '#99742A', '#CC8F1B', '#FFAA0D'],
+  teal: getColorRampByOpacitySteps('#00FFBC'),
```
nerik, line 49: 1. Are we sure we want those scales to be transparent? ie how does it plays on a satellite basemap for example?
2. Why is this better than hardcoded ramps that are not meant to be modified?
3. If this is the solution, why the whole hex2Rgb thing and not have directly values stored as rgb, ie `getColorRampByOpacitySteps(166,255,89)` instead of `getColorRampByOpacitySteps('#A6FF59')`
4. If really we prefer storing those values as hex, why do we need a `hex2Rgb` buried deep down in generators/heatmap/config instead of any readily available lib?

packages/layer-composer/src/generators/heatmap/heatmap.ts
```diff
@@ -65,8 +65,10 @@ class HeatmapGenerator {
 
     let stops: number[] = []
     const zoom = Math.min(Math.floor(config.zoom), config.maxZoom || HEATMAP_DEFAULT_MAX_ZOOM)
-    const statsByZoom = (this.stats !== null && this.stats[zoom]) || null
-    if (statsByZoom) {
+    const statsByZoom = (this.stats[config.id] && this.stats[config.id][zoom]) || null
+    if (config.steps) {
```
nerik, line 69: Should we rename these `breaks` ?

packages/layer-composer/src/generators/heatmap/heatmap.ts
```diff
@@ -158,7 +164,7 @@ class HeatmapGenerator {
 
     const promise = new Promise((resolve, reject) => {
       statsPromise.then((stats: statsByZoom) => {
-        this.stats = stats
+        this.stats[config.id] = stats
```
nerik, line 167: Why not use `memoizeByLayerId` ?

nerik (commented):
A few comments, feel free to ignore and postpone discussion to next monday

PR #86, opened 2020-09-18
initial map-legend hook

j8seangel:
Map legend hook to simplify style and dataviews and hover events consume


applications/amathea/src/features/map/Map.tsx
```diff
@@ -1,59 +1,42 @@
...
-  }, [currentValues, style])
-  return layersWithLegend
-}
-
 const Map = (): React.ReactElement => {
   const mapRef = useMapboxRef()
   const { viewport, onViewportChange, setMapCoordinates } = useViewport()
   const { latitude, longitude, zoom } = viewport
   const { logged } = useUserConnect()
+  const { dataviewsList } = useDataviewsConnect()
```
j8seangel, line 26: Now I'm having some doubts about needing this everywhere... why should map know about dataviews?
maybe it is easier to merge at the beginning of the chain, aka layer-composer, and grab every information from the `style` output but let's discuss next week

PR #90, opened 2020-09-25
New basemap

nerik:
Adds new basemap composed of bathymetry + landmass + graticules

Screenshot 2020-09-24 at 19 00 03

Full basemap for future dev including countries/regions/ports is stored here:
https://github.com/GlobalFishingWatch/basemap


j8seangel (approved):

packages/layer-composer/src/generators/types.ts
```diff
@@ -65,7 +65,7 @@ export interface BasemapGeneratorConfig extends GeneratorConfig {
   /**
    * Sets the color of the map background in any format supported by Mapbox GL, see https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#color
    */
-  basemap?: string
+  basemap: BasemapType
```
j8seangel, line 68: I think we could leave it as optional because we already have a default configuration

PR #94, opened 2020-09-29
fishing-map: Fourwings test

nerik:
Screenshot 2020-09-29 at 12 02 09

Test 4wings generator with timeline interaction.

### DIFF
--- a/applications/fishing-map/src/features/map/map.slice.ts
+++ b/applications/fishing-map/src/features/map/map.slice.ts
@@ -6,7 +6,7 @@ import {
   AnyGeneratorConfig,
   BackgroundGeneratorConfig,
   BasemapGeneratorConfig,
-  CartoPolygonsGeneratorConfig,
+  // CartoPolygonsGeneratorConfig,
   HeatmapAnimatedGeneratorConfig,
 } from '@globalfishingwatch/layer-composer/dist/generators/types'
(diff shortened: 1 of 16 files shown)


applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -18,9 +20,32 @@ const initialState: MapState = {
...
+      type: Generators.Type.HeatmapAnimated,
+      sublayers: [
+        { id: '0', colorRamp: 'teal', datasets: ['fishing_v4'], filter: "flag='ESP'" },
+        { id: '1', colorRamp: 'magenta', datasets: ['fishing_v4'], filter: "flag='FRA'" },
+      ],
+      combinationMode: 'compare',
+      debug: false,
+      debugLabels: false,
+      geomType: 'gridded',
+      tilesAPI: ' https://fourwings-tile-server-jzzp2ui3wq-uc.a.run.app/v1',
```
j8seangel, line 46: isn't this going through the gateway?

j8seangel (changes requested):
Looks good, but the build failed with the error:
`Object literal may only specify known properties, and 'datasets' does not exist in type 'HeatmapAnimatedGeneratorSublayer'` and I tested locally with the same error, does it work well for you? or what are we missing?

PR #107, opened 2020-10-02
Connect worskpace 4wings

j8seangel:
- Get all dataviews from workspace instead of map.slice to have a single source of truth.
- Include flag filters in 4wings `fishing_v4` layer

Also, it uses the workspace viewport when defined or a fallback in case there isn't in the workspace or the url.


applications/fishing-map/src/features/workspace/workspace.mock.ts
```diff
@@ -1,27 +1,55 @@
...
+        type: Generators.Type.HeatmapAnimated,
         color: '#00FFBC',
         colorRamp: 'teal',
+        sublayers: [{ id: 'fishing', colorRamp: 'teal', datasets: ['fishing_v4'] }],
+        combinationMode: 'compare',
+        debug: false,
+        debugLabels: false,
+        geomType: 'gridded',
+        tilesAPI: `${process.env.REACT_APP_API_GATEWAY}/v1/4wings`,
+        interactive: true,
```
nerik, line 52: I would remove non-mandatory generator params where we don't need another value than the default: debug, debugLabels, geomType, interactive

also `colorRamp` is deprecated.

nerik (commented):
LGTM 👍

PR #109, opened 2020-10-03
Resources slice

nerik:
Adds a simple Resources slice where resources are indexed by URL.
The URLs of resources are resolved by the `selectDataviewsResourceQueries` selector using the dataview, its dataset and its datasetParams.
 
Nevertheless, ATM the track URL is still hardcoded because I have no dataset id and no vessel id yet to use.

Also, I have changed AsyncReducerStatus to a TS enum, which I think which should start to use by default rather than loose `'foo' | 'bar'` string types.

### DIFF
--- a/applications/fishing-map/src/features/resources/resources.slice.ts
+++ b/applications/fishing-map/src/features/resources/resources.slice.ts
@@ -1,0 +1,69 @@
+import { createSlice, createAsyncThunk } from '@reduxjs/toolkit'
+import { AsyncReducerStatus } from 'types'
+import { RootState } from 'store'
+import { trackValueArrayToSegments, Field } from '@globalfishingwatch/data-transforms'
+import GFWAPI from '@globalfishingwatch/api-client'
+import { DataviewDatasetConfig, DatasetTypes } from '@globalfishingwatch/dataviews-client'
+
+export interface ResourceQuery {
+  url: string
+  dataviewId: number
+  datasetConfig: DataviewDatasetConfig
+  datasetType: DatasetTypes
+}
+
+export interface Resource extends ResourceQuery {
+  data: unknown
+}
+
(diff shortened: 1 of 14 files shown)


packages/dataviews-client/src/types.ts
```diff
@@ -10,6 +10,7 @@ export interface EndpointParam {
   enum?: string[]
   default?: string | boolean | number
   required?: boolean
+  values?: any
```
j8seangel, line 13: Let's have just enum or values as those do the same thing and seems to be duplicated, we need to agree with BE in which property we choose and remove one of them.

applications/fishing-map/src/features/resources/resources.slice.ts
```diff
@@ -0,0 +1,69 @@
...
+  data: unknown
+}
+
+interface ResourcesState {
+  status: AsyncReducerStatus
+  resources: Record<string, Resource>
+}
+
+const initialState: ResourcesState = {
+  status: AsyncReducerStatus.Idle,
```
j8seangel, line 25: I'd move this to the internal resource property as it is requested [in batch](https://github.com/GlobalFishingWatch/frontend/pull/109/files#diff-283a0939f4dbf6c8a3d9bf4ce071add1R53) will be nice to have the status of each request

j8seangel (approved):
Works and looks good! I'm going to merge for now to have it working on the dev environment but will need some tweaks later.
Thanks for the enum change, it is way better!

j8seangel:
Ah! I also included this [small change](https://github.com/GlobalFishingWatch/frontend/pull/109/commits/ed46a46927ab2682991286f4fda0acfabf052116) to be able to debug in redux the resource slice with only the raw data not serialized

PR #110, opened 2020-10-04
fishing map 4w click interaction

nerik:
Screenshot 2020-10-07 at 12 41 50

- leverages <a href="https://github.com/GlobalFishingWatch/mapbox-gl-js/pull/86">patch</a> done in MGL to get info about tile and 4wings cell when interacting
   - corresponding logic to retrieve that info in useMapClick hook
   - updates MGL to latest version in amathea, temporalgrid demo and fishing map (otherwise yarn install kept on installing an old version, we have a dependency version resolving issue)
- adds a basic react-map-gl popup component
- makes use of useMapClick and useMapTooltip hooks
- adds a slice to store hovered and click events
- adds an interaction endpoint mock
- adds logic to resolve that endpoint and attack it


applications/fishing-map/package.json
```diff
@@ -13,9 +13,9 @@
     "@globalfishingwatch/dataviews-client": "^3.0.0",
     "@globalfishingwatch/layer-composer": "^2.13.3",
     "@globalfishingwatch/linting": "^1.3.1",
-    "@globalfishingwatch/mapbox-gl": "1.12.0-gfw.1",
+    "@globalfishingwatch/mapbox-gl": "1.12.0-gfw.2",
     "@globalfishingwatch/react-hooks": "^3.0.0",
-    "@globalfishingwatch/react-map-gl": "^5.2.7-gfw.0",
+    "@globalfishingwatch/react-map-gl": "5.2.7-gfw.1-beta4",
```
j8seangel, line 18: If we had issues with mapbox-gl version, won't we have the same with react-map-gl? because amathea is using [5.2.7-gfw.0](https://github.com/GlobalFishingWatch/frontend/pull/110/files#diff-0dcf200b9aba8bb33308f5cb7290344cR19)

applications/fishing-map/src/features/map-features/map-features.hooks.ts
```diff
@@ -0,0 +1,136 @@
+import { useSelector, useDispatch } from 'react-redux'
```
j8seangel, line 1: why do we need `map` and `map-features`? what are the differences?

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -42,6 +42,14 @@ export const selectFishingDatasets = createSelector([selectFishingDataviews], (d
   return dataviews.flatMap((dataview) => getDatasetsByDataview(dataview))
 })
 
+export const selectTemporalgridDataviews = createSelector(
```
j8seangel, line 45: There is already a [selector to get this dataview](https://github.com/GlobalFishingWatch/frontend/blob/2b46157ea7b80f776ae7a700299e394a869988f5/applications/fishing-map/src/features/workspace/workspace.selectors.ts#L27), I agree this is a better name so we could remove the previous one and update the imports to use this one

applications/fishing-map/src/features/map-features/map-features.hooks.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+    if (!event || !event.features) return
+    dispatch(setClickedEvent(event))
+    // TODO should work for multiple features
+    const feature: ExtendedFeature = event.features[0]
+    if (!dataviews || !feature || !feature.temporalgrid) return
+
+    // TODO We assume here that temporalgrid dataviews appear in the same order as sublayers are set in the generator, ie indices will match feature.temporalgrid.sublayerIndex
+    const dataview = temporalgridDataviews[feature.temporalgrid.sublayerIndex]
+    // TODO How to get the proper id? Should be fishing_v4
```
j8seangel, line 37: good catch, we'll need to think deeply about this as I don't have an answer right now

j8seangel (commented):
It works well but seems we have lost the highlight cell feature we used to have, is this coming from not using the `useMapClick` hook?

applications/fishing-map/package.json
nerik: Honestly I don't really get it... With that setup Mapbox GL ends up with the correct versions in all instances in the `yarn.lock` file. 🤷‍♂️

applications/fishing-map/src/features/map-features/map-features.hooks.ts
nerik: `map-features` is for stuff that is overlaid on top of the map, not strictly necessary but that way we don't end up with gigantic map slice and map selectors and map component.

applications/fishing-map/src/features/workspace/workspace.selectors.ts
nerik: Ups, didn't see that, will fix

PR #118, opened 2020-10-07
Multiple vessel tracks

j8seangel:
Use multiple instances of dataviews in the workspace to have different vessel tracks
image

Something to be aware of is I had to change the [selectViewport selector ](https://github.com/GlobalFishingWatch/frontend/compare/multiple-vessel-tracks?expand=1#diff-1b8f6a9ef399f342292a815ccf77c742R21) from router.selectors to map.selectors to avoid circular dependencies. Because workspace.selectors depends on router.selector so router.selector couldn't include the `selectWorkspaceViewport` 

Also, some changes in amathea project to include any as is not compatible with Dataview types anymore, but we don't need to change it for now as the dataview in API is not updated


applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -4,6 +4,21 @@ import { Dataview, resolveEndpoint } from '@globalfishingwatch/dataviews-client'
 import { selectWorkspace } from 'features/workspace/workspace.slice'
 import { ResourceQuery } from 'features/resources/resources.slice'
 
+// TODO should come from search or 4wings cell - not sure how to get that when set in a workspace?
+const DATASET_ID = 'carriers-tracks:v20200507'
+
+export const getUniqueDataviewId = (dataview: Dataview) => {
```
nerik, line 11: I suggest also adding the generator type to the uid, which could help in debugging situations.

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -18,20 +33,54 @@ export const selectWorkspaceViewport = createSelector([selectWorkspace], (worksp
...
+          (wddc) =>
+            wddc.datasetId === datasetConfig.datasetId && wddc.endpoint === datasetConfig.endpoint
+        )
+        if (!workspaceDataviewDatasetConfig) return datasetConfig
+
+        return { ...datasetConfig, ...workspaceDataviewDatasetConfig }
+      })
+      const resolvedDataview = {
+        ...dataview,
+        uid: getUniqueDataviewId(dataview),
```
nerik: Why is that necessary since it's immediately rewritten?

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -18,20 +33,54 @@ export const selectWorkspaceViewport = createSelector([selectWorkspace], (worksp
   return workspace?.viewport
 })
 
-export const selectWorkspaceDataviews = createSelector([selectWorkspace], (workspace) => {
+export const selectWorkspaceDataviewsResolved = createSelector([selectWorkspace], (workspace) => {
   if (!workspace) return
-  // TODO resolve them here
-  return workspace.dataviews
+  // TODO move this to use-workspace helper
+  const dataviews = workspace.dataviews.flatMap((dataview) => {
```
nerik: I'm a bit puzzled by the logic... Why not a much simpler `workspaceDataviewsConfig.map` and then get the corresponding dataview for each `workspaceDataviewConfig`?
My idea was that a `workspaceDataviewConfig` would always be needed, not optional. So `workspace.dataviewsConfig` should have a `dataviewConfig` for the background and the basemap, event though it only uses the dataviews defaults.

applications/fishing-map/src/features/workspace/workspace.mock.ts
```diff
@@ -686,6 +686,20 @@ const workspace: Workspace = {
...
+        colorRamp: 'salmon',
+      },
+      datasetsConfig: [
+        {
+          datasetId: 'carriers-tracks:v20200507',
+          params: [{ id: 'vesselId', value: '0017a7b68-87c4-5af4-c46e-5bebee3ddceb' }],
+          endpoint: 'carriers-tracks',
+        },
+      ],
+    },
```
nerik, line 700:
```suggestion
    },
    {
      dataviewId: 10
    },
     {
      dataviewId: 11
    }
```

Add background and basemap dataviews instances, to avoid the end result being a weird mix of the `dataviews` and `dataviewsConfig`.
Simply put `dataviewsConfig`, is a flat list of everything in the workspace, without having to rely on any logic.

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -18,20 +34,78 @@ export const selectWorkspaceViewport = createSelector([selectWorkspace], (worksp
...
+      return {
+        ...dataview,
+        ...urlDataview,
+        config: {
+          ...dataview.config,
+          ...urlDataview?.config,
+        },
+      }
+    })
+    // Get dataviews defined uniquely in the url
```
nerik, line 89: Why would that ever happen?

applications/fishing-map/src/features/resources/resources.slice.ts
```diff
@@ -6,25 +6,21 @@ import GFWAPI from '@globalfishingwatch/api-client'
 import { DataviewDatasetConfig, DatasetTypes } from '@globalfishingwatch/dataviews-client'
 
 export interface ResourceQuery {
+  id: string
```
nerik, line 9: Why is that needed? IMHO we either use the URL as a unique id, or we use an id, but not both. 
Maybe I'm missing something but I'd rather call `resolveEndpoint` every time I need to get a URL to get a resource for a dataview, rather than having a double system.

nerik (changes requested):
I think we need to agree on some "philosophical" things. There's overrides overriding overrides logic here that seems to me too complex for our purposes. 
- there is always a dataviewsConfig present in the workspace for a dataview to show up 
- overrides in the URL always apply to an existing dataviewsConfig

Alos, I think the name `dataviewsConfig` is confusing as it is a different thing than `dataview.config` but with almost the same name.
How about `dataviewInstance` ?

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -18,20 +34,78 @@ export const selectWorkspaceViewport = createSelector([selectWorkspace], (worksp
   return workspace?.viewport
 })
 
-export const selectWorkspaceDataviews = createSelector([selectWorkspace], (workspace) => {
-  if (!workspace) return
-  // TODO resolve them here
-  return workspace.dataviews
-})
+export const selectWorkspaceDataviewsResolved = createSelector(
```
nerik, line 37: I'm a bit puzzled by the logic... Why not a much simpler workspaceDataviewsConfig.map and then get the corresponding dataview for each workspaceDataviewConfig?
My idea was that a workspaceDataviewConfig would always be needed, not optional. So workspace.dataviewsConfig should have a dataviewConfig for the background and the basemap, event though it only uses the dataviews defaults.

j8seangel: Totally agree, I mess up things instead of keeping it simple, I've included the dataviewConfig for the basemap and now it is much easier to loop only thought `workspaceDataviewConfig` see [this commit](https://github.com/GlobalFishingWatch/frontend/pull/118/commits/5b98d0ee7bb9914d49f0ecdf0e2c70ead87b009b)

applications/fishing-map/src/features/resources/resources.slice.ts
j8seangel: Good catch, I created a `resolveDataviewDatasetResource` to get the same url in different places of the app easily and removed the `id` see https://github.com/GlobalFishingWatch/frontend/pull/118/commits/de346d3f6e90567c076e9658922f83ced5e4aea5

applications/fishing-map/src/features/workspace/workspace.selectors.ts
j8seangel: This is not used anymore as we have a unique identifier for each dataviewConfig

applications/fishing-map/src/features/map/map.selectors.ts
```diff
@@ -28,48 +39,34 @@ export const selectGlobalGeneratorsConfig = createSelector(
...
+        data = resources[url].data
       }
 
       return {
         ...config,
         // TODO Add vessel id for tracks ie
         // dataview.datasetsConfig[?].query.find(q => q.id === 'id').value
-        id: `${config.type}_${dataview.id}_SOME_UNIQUE_ID`,
-        visible,
+        id: `${config.type}_${dataview.configId}`,
```
nerik, line 66: Actually, now that we have a proper uid generated (👍 ), I think this should simply be the dataviewInstance/dataviewConfig uid. 
Helps then to track stuff in style metadata.

applications/fishing-map/src/features/workspace/workspace.mock.ts
```diff
@@ -657,31 +658,41 @@ const workspace: Workspace = {
       ],
     },
   ],
+  // We need to "instanciate" every dataview we want to use in the workspace
+  // think as same model than mapbox-gl with source (dataview) and layer (dataviewConfig)
```
nerik, line 662: I strongly think this should be called `dataviewsInstances`

nerik (changes requested):
Looks much simpler now, great!
Please have a look at the two comments.

applications/fishing-map/src/features/workspace/workspace.mock.ts
j8seangel: ok, let's do it! but as this is something that becomes with libraries and API updates I'll do it in another PR

applications/fishing-map/src/features/map/map.selectors.ts
j8seangel: I agree but the `uid` is the [configId](https://github.com/GlobalFishingWatch/frontend/pull/118/files#diff-cdb6d9cab8c19208b3e9b07030066a29R85) now, see https://github.com/GlobalFishingWatch/frontend/pull/118/files#diff-975acf8e49c9a5d56577ed5bea82be8dR673.
It is mandatory when you add a dataview to a workspace instead of generated at runtime, or autogenerated when you add a new instance of the dataview to the workspace (for example, [adding a vessel from the search](https://github.com/GlobalFishingWatch/frontend/pull/118/files#diff-1d2e6e967661b4499f7c8472315e666aR54))

PR #120, opened 2020-10-08
Fishing map legend

nerik:
- generates color ramp legend using ui-component and react-portal
- made it possible for the legend component to be used as a standalone item (MapLegend/MapLegends)
- abstracted away breaks generation logic from Heatmap Generator and use it in HeatmapAnimated with hardcoded min/max/avg base
- applied multiplicator to breaks to fit time range

### DIFF
--- a/applications/fishing-map/src/features/map/Map.tsx
+++ b/applications/fishing-map/src/features/map/Map.tsx
@@ -1,17 +1,18 @@
 import React, { useCallback, useEffect, useState, useMemo } from 'react'
 import { AsyncReducerStatus } from 'types'
 import { createPortal } from 'react-dom'
+import { useSelector } from 'react-redux'
 import {
   MiniGlobe,
   IconButton,
   MiniglobeBounds,
   MapLegend,
-  LegendLayer,
 } from '@globalfishingwatch/ui-components/dist'
 import { InteractiveMap, ScaleControl, MapRequest } from '@globalfishingwatch/react-map-gl'
 import GFWAPI from '@globalfishingwatch/api-client'
 import { useLayerComposer, useMapClick } from '@globalfishingwatch/react-hooks'
 import { useClickedEventConnect, useMapTooltip } from 'features/map/map.hooks'
+import { selectWorkspaceDataviewsResolved } from 'features/workspace/workspace.selectors'
 import { ClickPopup } from 'features/map/Popup'
 import { useGeneratorsConnect } from './map.hooks'
 import useViewport from './map-viewport.hooks'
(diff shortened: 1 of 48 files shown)


packages/layer-composer/src/generators/heatmap/heatmap-animated.ts
```diff
@@ -88,6 +84,41 @@ const toURLArray = (paramName: string, arr: string[]) => {
...
+  const ramps = getSublayersColorRamps(config)
+  return breaks.map((sublayerBreaks, sublayerIndex) => {
+    const ramp = ramps[sublayerIndex]
+    const legendRamp = sublayerBreaks.map((break_, breakIndex) => {
+      // TODO Omitting the Zero value hence the +1
+      const rampColor = ramp[breakIndex + 1] as string
+      const legendRampItem: [number, string] = [break_, rampColor]
+      return legendRampItem
+    })
+    const sublayerLegend: LayerMetadataLegend = {
```
j8seangel, line 111: This needs to be defined out of the generator because will probably come from the dataview + dataset config... and I have two options in mind:
1. Prepare the data in `selectWorkspaceDataviewsResolved` and pass it down to the generator even it won't use it but seems easier as it contains all the legend data
2. Pass the dataviewInstance id to the generator and get it back to merge them in the app as you did [here](https://github.com/GlobalFishingWatch/frontend/pull/120/files#diff-1970e0dbbe9781cbaeda038430e2e38eR100)

I think the first one is easier but what are your thoughts ?

applications/fishing-map/src/features/map/Map.tsx
```diff
@@ -115,6 +146,11 @@ const Map = (): React.ReactElement => {
         <IconButton icon="ruler" type="map-tool" tooltip="Ruler (Coming soon)" />
         <IconButton icon="camera" type="map-tool" tooltip="Capture (Coming soon)" />
       </div>
+      {layersWithLegend?.map(
+        (legend) =>
+          document.getElementById(legend.id) &&
+          createPortal(<MapLegend layer={legend} />, document.getElementById(legend.id) as Element)
```
j8seangel, line 152: This is a very good idea! simplify a lot the needs of sharing the derived styled state! 👏

applications/fishing-map/src/features/sidebar/heatmaps/HeatmapLayerPanel.tsx
```diff
@@ -97,6 +97,9 @@ function LayerPanel({ dataview }: LayerPanelProps): React.ReactElement {
       <div className={styles.expandedContainer} ref={expandedContainerRef}>
         {filterOpen && <Filters dataview={dataview} />}
       </div>
+      {/* TODO Use the real dataview instance id */}
```
j8seangel, line 100: I will update this with the unique dataviewInstance id

j8seangel (commented):
Looks good! there is an open conversation about how to handle legend metadata from dataview, see comment below, but this a very good starting point!

packages/layer-composer/src/generators/heatmap/heatmap-animated.ts
nerik: Good catch, and definitely 2., which was easy to do

PR #124, opened 2020-10-13
Vessel resource info

j8seangel:
Huge PR coming... sorry but it had a lot of implications:
- Rename dataviewConfig to dataviewInstance
- Move workspace dataviews and datasets to its own slice using generic async-slice
- Replace every DATASET_ID usage in order to find by TYPE instead and remove fixed versions
- Search vessels by related datasets instead of fixed dataset version https://github.com/GlobalFishingWatch/frontend/compare/vessel-resource-info?expand=1#diff-6538b758ec433f59cf22639d66507de5674d4ac072b37cddb53fc5577f2df16fR114
- Had to [embed the dataset](https://github.com/GlobalFishingWatch/frontend/compare/vessel-resource-info?expand=1#diff-b2c98b25aadabf8c2ec0693cd6474baba811240dac28fea467cfaba667c11049R29) used in the interaction to retrieve the relatetDatasets for the vessel info but I don't really like the option, any thought here?


applications/fishing-map/src/features/map/Popup.tsx
```diff
@@ -23,20 +23,25 @@ function PopupWrapper({
...
-        },
-      ],
+  const { upsertDataviewInstance } = useDataviewInstancesConnect()
+  const onVesselClick = (vessel: any, feature: TooltipEventFeature) => {
+    const datasetId = feature.dataset?.relatedDatasets?.find(
+      (relatedDataset) => relatedDataset.type === TRACKS_DATASET_TYPE
+    )?.id
+    if (datasetId) {
+      const dataviewInstance: DataviewInstance = {
+        id: `track-${vessel.id}`,
```
nerik, line 33: The logic to generate this should be centralized somewhere, duplicate with Search.tsx 
https://github.com/GlobalFishingWatch/frontend/pull/124/files#diff-0fc787fc5bdbf9885682179d3112abad8749adc32b96fd9b5f511b878b7c089eR70

applications/fishing-map/src/features/workspace/workspace.hook.ts
```diff
@@ -1,50 +1,52 @@
...
+export const useDataviewInstancesConnect = () => {
+  const urlDataviewInstances = useSelector(selectDataviewInstances)
+  const workspaceDataviewInstances = useSelector(selectWorkspaceDataviewInstances)
   const { dispatchQueryParams } = useLocationConnect()
 
-  const updateDataviewConfig = useCallback(
-    (dataviewConfig: Partial<WorkspaceDataviewConfig>) => {
-      const currentDataviewConfig = urlDataviewsConfig?.find(
-        (urlDataviewConfig) => urlDataviewConfig.id === dataviewConfig.id
+  const upsertDataviewInstance = useCallback(
```
nerik, line 13: I'm torn about this. It's looks like a smart way to add or update an instance, but I also feel that you always know whether you're adding or updating (ie adding if loading track for search, updating if toggling visibility in HeatmapLayerPanel for instance).
I am worried we'll be doing by error `add` instead of `update` because somehow the id wrong.

- export two distinct explicit methods?
- or maybe throw an error if `id` is provided and dataviewInstance found in `urlDataviewInstances` (in which case the upsert should be responsible for generating an id when doing `add`, which might be a good idea, see my other comment)

packages/dataviews-client/src/dataviews-client.ts
```diff
@@ -1,7 +1,7 @@
 import template from 'lodash/template'
 import { stringify } from 'qs'
 import GFWAPI, { FetchResponseTypes, FetchOptions } from '@globalfishingwatch/api-client'
-import { Dataview, Resource, WorkspaceDataviewConfig } from './types'
+import { Dataview, Resource, DataviewInstance } from './types'
```
nerik, line 4: I think you modified that just doing a search & replace, but I think we could get rid of this file and think about where to put `resolveEndpoint` and the types.

nerik (changes requested):
Looks awesome, we are getting closer to SANITY

applications/fishing-map/src/features/map/Popup.tsx
j8seangel: It isn't easy as we have to configure the dataset for the track and the same for the info with different endpoints for each one... I've created a `getVesselDataviewInstance` util in dataviews to centralize but will need some review later

packages/dataviews-client/src/dataviews-client.ts
j8seangel: Good idea, I'll move in another PR this and some other several types to a new package to include only types

applications/fishing-map/src/features/workspace/workspace.hook.ts
j8seangel: I'm not sure I got your point here, but some thoughts here:

About your first comment, I could create two separate methods to be able to use them explicitly when needed but then we'll need to duplicate logic to check first if it already exists in the URL.

Then, your worry will be even bigger as the logic of checking when you need to add or update is distributed on different components as this won't solve the id error because the id has always to be provided.

PR #130, opened 2020-10-15
Vessel popup info

j8seangel:
Request vessel details info in 4wings popup cell with an important improvement as now the request is [canceled](https://redux-toolkit.js.org/api/createAsyncThunk#canceling-while-running) when clicked in another cell to avoid race conditions


applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -29,12 +34,50 @@ type Fetch4WingInteractionThunk = {
 }
 export const fetch4WingInteractionThunk = createAsyncThunk(
   'map/fetchInteraction',
-  async ({ dataset, datasetConfig }: Fetch4WingInteractionThunk) => {
+  async (
+    { dataset, datasetConfig }: Fetch4WingInteractionThunk,
+    { getState, signal, rejectWithValue }
```
nerik, line 39: `rejectWithValue` is not used

applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -29,12 +34,50 @@ type Fetch4WingInteractionThunk = {
...
+              {
+                id: 'ids',
+                value: vesselsForDataset.slice(0, MAX_TOOLTIP_VESSELS).map((v) => v.id),
+              },
+            ],
+          }
+          const infoUrl = resolveEndpoint(infoDataset, infoDatasetConfig)
+          if (infoUrl) {
+            // TODO create search API results response
+            const vesselsInfo = await GFWAPI.fetch<any>(infoUrl, { signal })
```
nerik, line 68: What's `signal` for?

nerik (commented):
Looks good!
One thing I don't understand is that when info endpoint fails to load (in fact it never works for me, always getting 403s), the popup doesn't even populate with just vessel ids, it's empty. Maybe we should have a generic "unknown vessel" that's still clickable?

applications/fishing-map/src/features/map/map.slice.ts
j8seangel: it is an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal) instance property that can be used to abort a fetch request

applications/fishing-map/src/features/datasets/datasets.mock.ts
```diff
@@ -82,6 +82,33 @@ const datasets: Dataset[] = [
...
+            required: true,
+          },
+          {
+            label: 'Ids',
+            id: 'ids',
+            type: 'string',
+            required: true,
+          },
+          {
+            label: 'binary',
```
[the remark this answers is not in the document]
j8seangel: I've just checked it out and seems that  it could be also requested as binary to save some network data but I couldn't found the pbf decoder, I'll ask API team and, thanks for the catch!

applications/fishing-map/src/features/map/Map.tsx
```diff
@@ -110,9 +110,9 @@ const MapWrapper = (): React.ReactElement => {
         ? layer.metadata.legend
         : [layer.metadata.legend]
 
+      // TODO Move this to its own package
```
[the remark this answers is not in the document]
j8seangel: Indeeed

j8seangel:
Thanks for the review, now the vessels info request error is caught so we will always see something in the modal

PR #133, opened 2020-10-16
Merge 4w sublayers 

nerik:
- Sets up mock workspace with two 4w based dataviews
- Logic to merge all 4w based dataviews into a single generator with multiple sublayers
- Store filters in dataview instances 

Screenshot 2020-10-16 at 12 47 25


j8seangel (approved):
🥇

PR #134, opened 2020-10-16
Workspace id path

j8seangel:
Prepare routes path to support, for example, `/indonesia` using indonesia as workspace id.
Also supports API version path with `/:workspaceId/v1|v2...`

I had to remove the `/search` path as was hard to make it compatible with the workspaceId and version optional params, now is used with the `query` queryParam and also moved the sidebarOpen state to URL too

Something to notice is an easy improvement using [redux-toolkit batch](https://react-redux.js.org/api/batch) [here](https://github.com/GlobalFishingWatch/frontend/compare/workspace-id-path?expand=1#diff-0fc787fc5bdbf9885682179d3112abad8749adc32b96fd9b5f511b878b7c089eR38) to group actions dispatch and avoid multiple re-renders

### DIFF
--- a/applications/fishing-map/src/data/config.ts
+++ b/applications/fishing-map/src/data/config.ts
@@ -7,10 +7,13 @@ export const FALLBACK_VIEWPORT = {
   zoom: 3,
 }
 
+export const DEFAULT_WERSION = 'v1'
+
 export const DEFAULT_WORKSPACE = {
   latitude: undefined,
   longitude: undefined,
   zoom: undefined,
+  query: undefined,
   start: new Date(2019, 0, 1).toISOString(),
   end: new Date(2019, 1, 1).toISOString(),
   availableStart: new Date(2012, 0, 1).toISOString(),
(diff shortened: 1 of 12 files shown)


applications/fishing-map/src/features/search/Search.tsx
```diff
@@ -23,29 +23,31 @@ import SearchEmptyState from './SearchEmptyState'
...
 
   useEffect(() => {
-    if (query) {
-      dispatchLocation(SEARCH, { query })
-      dispatch(fetchVesselSearchThunk({ query, datasets: searchDatasets }))
-    } else {
-      dispatchLocation(SEARCH)
-      dispatch(cleanVesselSearchResults())
+    if (query !== '') {
+      batch(() => {
```
nerik, line 38: That's pretty cool.

applications/fishing-map/src/data/config.ts
```diff
@@ -7,10 +7,14 @@ export const FALLBACK_VIEWPORT = {
   zoom: 3,
 }
 
+export const DEFAULT_WERSION = 'v1'
```
nerik, line 10: I weawy wike this new wersion

nerik (commented):
Vewy nice, LGTM

PR #136, opened 2020-10-16
CSS for hover and click popups

satellitestudiodesign:


nerik (commented):
Looking good!
I took the liberty of using MGL native `Anchor` type.

Also:
- spinner doesn't fit in popup:
Screenshot 2020-10-16 at 14 50 22
- we will need to have a look at how we name 4w sublayers, because that's not ideal:
Screenshot 2020-10-16 at 14 49 36
- popup subsection shouldn't show up when having a value of 0 hours, it's on me 
(adding last 2 to backlog)

PR #139, opened 2020-10-19
Api types packages

j8seangel:
Create a new package called `api-types` to unify API existing response types and update their use in:
- [x] other packages
- [x] fishing-map
- [x] amathea
- [x] sandbox

as temporal-grid is deprecated for now, there hasn't been updated in this PR.

It also fixes tracks and use proto buffer to save some network transfer data


nerik (approved):
Much cleaner, thanks!

PR #146, opened 2020-10-19
protected context area

j8seangel:
MPA as context area example:
image

Also, restore dataviewInstance on `+` icon button click in context areas and heatmap after removal all of them

### DIFF
--- a/applications/fishing-map/src/features/sidebar/context-areas/ContextAreaLayerPanel.tsx
+++ b/applications/fishing-map/src/features/sidebar/context-areas/ContextAreaLayerPanel.tsx
@@ -35,7 +35,7 @@ function LayerPanel({ dataview }: LayerPanelProps): React.ReactElement {
   }
 
   const color = dataview?.config?.color
-  const changeTrackColor = (color: ColorBarOption) => {
+  const changeColor = (color: ColorBarOption) => {
     upsertDataviewInstance({
       id: dataview.id,
       config: { color: color.value },
@@ -111,7 +111,7 @@ function LayerPanel({ dataview }: LayerPanelProps): React.ReactElement {
           <ColorBar
             colorBarOptions={TrackColorBarOptions}
             selectedColor={dataview.config?.color}
-            onColorClick={changeTrackColor}
+            onColorClick={changeColor}
           />
         )}
       </div>
--- a/applications/fishing-map/src/features/sidebar/context-areas/ContextAreaLayerPanel.tsx
+++ b/applications/fishing-map/src/features/sidebar/context-areas/ContextAreaLayerPanel.tsx
@@ -21,7 +21,7 @@ type LayerPanelProps = {
 }
 
 function LayerPanel({ dataview }: LayerPanelProps): React.ReactElement {
-  const { upsertDataviewInstance, deleteDataviewInstance } = useDataviewInstancesConnect()
+  const { upsertDataviewInstance } = useDataviewInstancesConnect()
   const { url } = resolveDataviewDatasetResource(dataview, VESSELS_DATASET_TYPE)
   const resource = useSelector(selectResourceByUrl<Vessel>(url))
   const [colorOpen, setColorOpen] = useState(false)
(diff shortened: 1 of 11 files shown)


applications/fishing-map/src/features/sidebar/context-areas/ContextAreaLayerPanel.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const layerActive = dataview?.config?.visible ?? true
+  const onToggleLayerActive = () => {
+    upsertDataviewInstance({
+      id: dataview.id,
+      config: { visible: !layerActive, color },
+    })
+  }
+
+  const color = dataview?.config?.color
+  const changeTrackColor = (color: ColorBarOption) => {
```
nerik, line 38: nitpick
```suggestion
  const changeColor = (color: ColorBarOption) => {
```

applications/fishing-map/src/features/sidebar/context-areas/ContextAreaSection.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { selectContextAreasDataviews } from 'features/workspace/workspace.selectors'
+import styles from 'features/sidebar/common/Sections.module.css'
+import LayerPanel from './ContextAreaLayerPanel'
+
+function VesselsSection(): React.ReactElement {
```
nerik, line 7: nitpick
```suggestion
export default function(): React.ReactElement {
```

applications/fishing-map/src/features/sidebar/heatmaps/HeatmapsSection.tsx
```diff
@@ -1,17 +1,29 @@
...
 import { selectTemporalgridDataviews } from 'features/workspace/workspace.selectors'
 import styles from 'features/sidebar/common/Sections.module.css'
+import { useDataviewInstancesConnect } from 'features/workspace/workspace.hook'
 import LayerPanel from './HeatmapLayerPanel'
 
 function HeatmapsSection(): React.ReactElement {
   const dataviews = useSelector(selectTemporalgridDataviews)
+  const { removeDataviewInstance } = useDataviewInstancesConnect()
+  const onAddClick = useCallback(() => {
+    removeDataviewInstance('fishing-1')
```
nerik, line 13: Unclear why `onAddSomething` triggers `removeSomething`.
Maybe `toggleDataviewInstance` ?

nerik (changes requested):
Looks good 👍 
I have a bunch of comments about naming. Sometimes maybe coming from some copy-pasted code. It's just nitpick but probably better to fix that now to avoid future WTF moments.
In the same spirit: the UserContext generator is deifinitely a good base for context layers, it would make sense to rename it simply Context. Adding that to the backlog + add interaction

applications/fishing-map/src/features/sidebar/heatmaps/HeatmapsSection.tsx
j8seangel: For now, this is just a workaround that will be deleted later but something we need to discuss is how to remove an existing dataviewInstance from the workspace...

j8seangel:
Thanks for the naming catches, I agree with all of them except in:

> In the same spirit: the UserContext generator is definitely a good base for context layers, it would make sense to rename it simply Context. Adding that to the backlog + add interaction

it is done because this was the way to have it working easily but we'll need to keep user-context as simple as it is, for now, and create a new one with interaction, different styles for the EEZ, etc... but added a TODO to not forget it

applications/fishing-map/src/features/sidebar/context-areas/ContextAreaSection.tsx
j8seangel: I prefer having the name so in the Component debugger tree it is easily to debug rathern than `anonymous function`

nerik: I was bothered by the fact it's called 'Vessels'

applications/fishing-map/src/features/sidebar/heatmaps/HeatmapsSection.tsx
nerik: Please add a comment/TODO.

PR #152, opened 2020-10-20
adds UI and queryParams to set Timebar settings

satellitestudiodesign:
Screenshot 2020-10-20 at 16 03 26
😬


applications/fishing-map/src/data/config.ts
```diff
@@ -46,3 +49,45 @@ export const sources: SelectOption[] = [
 ]
 
 export const DATE_FORMAT = 'yyyy/MM/dd'
+
+export const timebarEventsOptions: SelectOption[] = [
```
nerik, line 53:
```suggestion
export const TIMEBAR_EVENT_OPTIONS: SelectOption[] = [
```
Please use SCREAMING_SNAKE_CASE by convention when setting readonly consts

applications/fishing-map/src/data/config.ts
```diff
@@ -46,3 +49,45 @@ export const sources: SelectOption[] = [
...
+    id: 'ports',
+    label: 'Port visits',
+  },
+  {
+    id: 'none',
+    label: 'None',
+  },
+]
+
+export const timebarGraphOptions: SelectOption[] = [
```
nerik, line 80:
```suggestion
export const TIMEBAR_GRAPH_OPTIONS: SelectOption[] = [
```


Please use SCREAMING_SNAKE_CASE by convention when setting readonly consts

applications/fishing-map/src/types/index.ts
```diff
@@ -40,3 +46,7 @@ export enum AsyncReducerStatus {
   Finished = 'finished',
   Error = 'error',
 }
+
+export type TimebarVisualisations = 'heatmap' | 'vessel'
+export type TimebarEvents = 'all' | 'fishing' | 'encounters' | 'loitering' | 'ports' | 'none'
+export type TimebarGraphs = 'speed' | 'depth' | 'none'
```
nerik, line 52: (nitpick) I prefer enums than magic strings, but not important.
```suggestion
export enum TimebarGraphs {
  Speed = 'speed'
  Depth = 'depth,
  None = 'none',
}

```

(then use `TimebarGraphs.Speed` instead of 'speed', which will give you autocomplete)

applications/fishing-map/src/features/timebar/TimebarSettings.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const removeEventsOption = () => {
+    dispatchQueryParams({ timebarEvents: 'none' })
+  }
+  const removeGraphOption = () => {
+    dispatchQueryParams({ timebarGraph: 'none' })
+  }
+  const expandedContainerRef = useClickedOutside(closeOptions)
+
+  if (activeHeatmapDataviews?.length === 0 && timebarVisualisation === 'heatmap')
+    dispatchQueryParams({ timebarVisualisation: 'vessel' })
```
nerik: There are problems with this:
- it might spam the whole app by changing the query params whenever this component is rendered
(having imperative code that affects the whole app state in a component render is not a good idea)
- you are not actually keeping the user preference when defaulting to vessel

See another approach here https://github.com/GlobalFishingWatch/frontend/pull/152/commits/601c3d2729afb0a690e94085e26e83f7b191ac7d
(you could also do that in a selector)

nerik (changes requested):
Looks good!

satellitestudiodesign:
thanks

PR #153, opened 2020-10-21
Fishingmap/screenshot

j8seangel:
- [x] Move all controls to its own folder trying to keep it clean
- [x] Modal with screenshot preview and download option:
image


nerik (changes requested):
LGTM appart from this:
It's unclear to me when hooks go to `src/hooks` and when they go to `features/*`. Can we decide on one? (for me strong preference on `features/*` as you already know)
Can we at least not have two files named `map.hooks.ts` ?

j8seangel:
> Can we at least not have two files named map.hooks.ts ? 

for sure, sorry I forgot the existing one in below features and I've moved it

It's unclear to me when hooks go to src/hooks and when they go to features/*

When something is generic for me fits in `hooks` like sizing elements, clicking outside, etc.. but if you prefer to create a feature for these kinds of things is the same for me

nerik:
@j8seangel Thanks for the change

> When something is generic for me fits in hooks like sizing elements, clicking outside, etc..

Yeah ok, makes sense. It was not obvious with `map.hooks`, makes sense for click outside and screen, I should probably move the debug menu one to app/

j8seangel:
> I should probably move the debug menu one to app 

Ok, I'm going to create an `app` feature and move it there

PR #166, opened 2020-10-25
added french option in debug menu

nerik:


packages/ui-components/src/switch/Switch.tsx
```diff
@@ -1,14 +1,19 @@
-import React, { memo } from 'react'
+import React, { MouseEvent, memo } from 'react'
 import cx from 'classnames'
 import { Placement } from 'tippy.js'
 import Tooltip from '../tooltip'
 import { TooltipTypes } from '../types/index'
 import styles from './Switch.module.css'
 
+// TODO Maybe a simple way is to have the Switch component wrap an <input type="checkbox"> so that we can use the React native event
```
j8seangel, line 8: It is a controlled component, so the active prop would be already available on the parent component but I agree this is a good practice

PR #174, opened 2020-10-27
[WIP] nohoist test

nerik:
Getting the React rule of hooks error when using that :/


j8seangel:
I couldn't fix it either, but as I updated react dependencies and fix the Amathea build I'm going to merge this and take a look at it later

PR #176, opened 2020-10-27
useTileState

nerik:
Trying to do a useTileState hook and failing :(
(seems to enter an infinite render loop)


j8seangel:
I've moved the hook to the `react-hooks` package fixing:
- the infinite loop (using [setState as function](https://reactjs.org/docs/hooks-reference.html#functional-updates) to avoid re-run the code on every current state change)
- the count avoid mutating the object

and also included in the fishing-map, for now using only the `loading` property with some changes to simplify:
- new useMapRefReady hook to ensure we only attach listeners once
- merge Map and MapWrapper components as had a lot of dependencies in common and added more complexity
- externalize some functions to keep the Map component clear

PR #186, opened 2020-11-04
Fishingmap/legend label

j8seangel:
Support custom label component to be able to render in this case only unit and gridArea with custom styles
ColorRamp
image
Bivariate:
image

⚠️ To avoid circular dependencies building I've removed the api-types changes watch so packages build should be as fast as before

### DIFF
--- a/applications/fishing-map/src/features/i18n/i18nNumber.tsx
+++ b/applications/fishing-map/src/features/i18n/i18nNumber.tsx
@@ -1,14 +1,33 @@
 import React, { Fragment } from 'react'
 import { useTranslation } from 'react-i18next'
 import { LocaleOptions } from 'types'
+import i18n from './i18n'
 
 type I18Number = string | number
+type I18NumberOptions =
+  | LocaleOptions
+  | (Intl.NumberFormatOptions & {
+      locale?: LocaleOptions
+      unit?: string
+      unitDisplay?: 'long' | 'short' | 'narrow'
+      notation?: 'standard' | 'scientific' | 'engineering' | 'compact'
+    })
 
-export const formatI18nNumber = (num: I18Number, locale: LocaleOptions) => {
-  const number = num === 'string' ? parseFloat(num) : num
-  return number.toLocaleString(locale, {
-    maximumFractionDigits: number < 10 ? 2 : 0,
-  })
+export const formatI18nNumber = (
+  number: I18Number,
+  options: I18NumberOptions = i18n.language as LocaleOptions
+) => {
+  const locale = typeof options === 'object' ? options.locale || i18n.language : options
+  const parsedNumber = number === 'string' ? parseFloat(number) : (number as number)
+  try {
+    return new Intl.NumberFormat(locale, {
+      maximumFractionDigits: number < 10 ? 2 : 0,
+      ...(typeof options === 'object' && { ...options }),
+    }).format(parsedNumber)
+  } catch (e) {
+    console.warn(e)
+    return number
+  }
 }
 
 export const useI18nNumber = (number: I18Number) => {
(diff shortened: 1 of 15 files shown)


applications/fishing-map/src/features/map/Map.tsx
```diff
@@ -211,18 +199,36 @@ const MapWrapper = (): React.ReactElement | null => {
...
-        (legend) =>
-          document.getElementById(legend.id as string) &&
-          createPortal(
+      {layersWithLegend?.map((legend) => {
+        const legendDomElement = document.getElementById(legend.id as string)
+        if (legendDomElement) {
+          const isSquareKm = (legend.gridArea as number) > 50000
+          const gridArea = isSquareKm ? (legend.gridArea as number) / 1000000 : legend.gridArea
+          const gridAreaFormatted = gridArea
+            ? formatI18nNumber(gridArea, {
```
nerik, line 208: Any reason for not putting that logic inside `getLegendLayers` ? This seems pretty generic

packages/layer-composer/src/generators/heatmap/config.ts
```diff
@@ -26,6 +26,22 @@ export const HEATMAP_MODE_LAYER_TYPE: Record<HeatmapAnimatedMode, string> = {
   [HeatmapAnimatedMode.Extruded]: 'fill-extrusion',
 }
 
+export const gridAreaByZoomLevel = [
```
nerik, line 29:
```suggestion
export const GRID_AREA_BY_ZOOM_LEVEL = [
```

applications/fishing-map/src/features/i18n/i18nNumber.tsx
```diff
@@ -1,14 +1,33 @@
 import React, { Fragment } from 'react'
 import { useTranslation } from 'react-i18next'
 import { LocaleOptions } from 'types'
+import i18n from './i18n'
 
 type I18Number = string | number
+type I18NumberOptions =
+  | LocaleOptions
```
nerik, line 8: Not related with PR sorry, but do you think we could rename `LocaleOptions` to `Locale` ?
`LocaleOptions` makes me think this an object with options for a locale.

applications/fishing-map/src/features/i18n/i18nNumber.tsx
```diff
@@ -1,14 +1,33 @@
...
-    maximumFractionDigits: number < 10 ? 2 : 0,
-  })
+export const formatI18nNumber = (
+  number: I18Number,
+  options: I18NumberOptions = i18n.language as LocaleOptions
+) => {
+  const locale = typeof options === 'object' ? options.locale || i18n.language : options
+  const parsedNumber = number === 'string' ? parseFloat(number) : (number as number)
+  try {
+    return new Intl.NumberFormat(locale, {
```
nerik, line 23: This is so cool! And support is good in basically every possible browser https://caniuse.com/?search=Intl.NumberFormat

nerik (commented):
Perfect, thanks!

applications/fishing-map/src/features/map/Map.tsx
j8seangel: You are right, the parsing function should be there and in the component just the custom styles.

applications/fishing-map/src/features/i18n/i18nNumber.tsx
j8seangel: Count with it!

applications/fishing-map/src/features/i18n/i18nNumber.tsx
j8seangel: Yes! I didn't know it and it was surprising, really nice. Thanks to bring it!

PR #194, opened 2020-11-06
Contextual layers

j8seangel:
New context LC generator to support EEZ with conflicted lines

image

TODO: interaction


applications/fishing-map/src/features/dataviews/dataviews.utils.ts
```diff
@@ -1,12 +1,13 @@
...
 import { TrackColorBarOptions } from '@globalfishingwatch/ui-components/dist/color-bar'
+import { Generators } from '@globalfishingwatch/layer-composer'
 import { DEFAULT_VESSEL_DATAVIEW_ID } from 'data/datasets'
 
 export const getVesselDataviewInstance = (
   vessel: { id: string },
   trackDatasets: Dataset[],
   infoDatasets: Dataset[]
-): DataviewInstance => {
+): DataviewInstance<Generators.Type> => {
```
nerik, line 10: I like that 👍

applications/fishing-map/src/features/dataviews/dataviews.slice.ts
```diff
@@ -6,13 +6,18 @@ import { AsyncReducer, createAsyncSlice } from 'utils/async-slice'
 import { RootState } from 'store'
 import { getDatasetByDataview } from 'features/workspace/workspace.slice'
 import { fetchDatasetsByIdsThunk } from 'features/datasets/datasets.slice'
+import mockedDataviews from './dataviews.mock'
```
nerik, line 9: I'm not sure having a mock file around that is only for "personal" use is a good idea. How about we have a `./dataviews.mock.sample` in git, but the actual file is not, and import the actual file when an env var is set to true?

applications/fishing-map/src/features/map/map.selectors.ts
```diff
@@ -111,40 +110,49 @@ export const getGeneratorsConfig = createSelector(
...
+        if (Array.isArray(dataview.config.layers)) {
+          const tilesUrls = dataview.config.layers?.flatMap(({ id, dataset }) => {
+            const { url } = resolveDataviewDatasetResource(dataview, { id: dataset })
+            if (!url) return []
+            return { id, tilesUrl: url }
+          })
+          // Duplicated generators when context dataview have multiple layers
+          return tilesUrls.map(({ id, tilesUrl }) => ({
+            ...generator,
+            id: dataview.id + id,
```
nerik, line 140: Shouldn't w avoid concatenating strings with `+` ? We used to have an eslint rule for that no?

nerik: (personally I prefer `[dataview.id, id].join('_')`

applications/fishing-map/src/features/workspace/workspace.default.ts
```diff
@@ -33,42 +33,51 @@ const workspace: Workspace = {
...
-              id: 'vesselId',
-              value: '00ba29183-3b86-9e36-cf20-ee340e409521',
-            },
-          ],
-          endpoint: 'carriers-vessel',
-          datasetId: 'test-alias-import-vessels-pipeline:latest',
-        },
-      ],
-    },
+    // {
```
nerik, line 36: Same comment as with `dataview.mock`

packages/api-types/package.json
```diff
@@ -17,14 +17,11 @@
     "url": "git+https://github.com/GlobalFishingWatch/frontend.git"
   },
   "scripts": {
-    "start": "rollup -c ./rollup.config.js",
+    "start": "rollup -c ./rollup.config.js -w",
```
nerik, line 20: 🙌

applications/fishing-map/src/types/index.ts
```diff
@@ -1,12 +1,13 @@
...
+import { Generators } from '@globalfishingwatch/layer-composer'
 
 export enum Locale {
   en = 'en',
   es = 'es',
   fr = 'fr',
 }
 
-export type UrlDataviewInstance = Omit<DataviewInstance, 'dataviewId'> & {
+export type UrlDataviewInstance = Omit<DataviewInstance<Generators.Type>, 'dataviewId'> & {
```
nerik, line 10: I like that too

packages/layer-composer/src/generators/context/context.ts
```diff
@@ -0,0 +1,134 @@
...
+    //   paint[`circle-color`] = [...matchFilter, circleColor, color]
+    //   paint['circle-opacity'] = [...matchFilter, circleOpacity, opacity]
+    //   paint['circle-stroke-color'] = [...matchFilter, circleStrokeColor, strokeColor]
+    //   paint['circle-stroke-width'] = [...matchFilter, circleStrokeWidth, strokeWidth]
+    // }
+    return circlePaint
+  }
+}
+
+class UserContextGenerator {
```
nerik, line 76:
```suggestion
class ContextGenerator {
```

packages/layer-composer/src/generators/context/context-layers.ts
```diff
@@ -0,0 +1,61 @@
...
+    },
+  ],
+  'eez-boundaries': [
+    {
+      id: 'eez_rest_lines',
+      type: 'line',
+      filter: ['match', ['get', 'line_type'], settledBoundaries, true, false],
+      layout: {},
+      paint: {
+        'line-color': '#93c96c',
```
nerik, line 37: Won't that value be overriden by `config.color` ?

packages/layer-composer/src/generators/types.ts
```diff
@@ -105,6 +107,25 @@ export interface UserContextGeneratorConfig extends GeneratorConfig {
   pickValueAt?: string
 }
 
+/**
+ * Contextual layers provided by GFW
+ */
+export interface ContextGeneratorConfig extends GeneratorConfig {
+  type: Type.Context
+  /**
+   * Sets the color of the line https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#paint-fill-fill-color
```
nerik, line 116: This comment doesn't belong here

packages/layer-composer/src/types/index.ts
```diff
@@ -1,5 +1,18 @@
...
+import type { DataviewConfig } from '@globalfishingwatch/api-types'
+import {
+  GeneratorConfig,
+  Type,
+  GeneratorLegend,
+  HeatmapAnimatedGeneratorSublayer,
+  ColorRampsIds,
+} from '../generators/types'
+
+export interface GeneratorDataviewConfig extends DataviewConfig<Type> {
```
nerik, line 11: I'm mot super clear why this type is needed, but isn't it a bit confusing? If we need a `HeatmapAnimatedConfig` or a `BasemapConfig`, wouldn't it be better to cast as one of these types rather than having this weird generic type?

nerik (changes requested):
I really like the "split into 2 generators" solution, very elegant 👍 
Please rebase before making changes.

packages/layer-composer/src/generators/context/context-layers.ts
j8seangel: It will, but there are some apps where we can't customize the color (for example, the carrier portal), and I think it is a good idea to have a color to render the layer equally in every app by default

applications/fishing-map/src/features/map/map.selectors.ts
j8seangel: Bought 💰

packages/layer-composer/src/types/index.ts
j8seangel: You are right, will be nice to have the option of typing by each different one and we'll use the one that composes all of them as fallback
So it will be Type by default but in the instance, we could pass Heatmap or Basemap or whatever we need.
```suggestion
export interface GeneratorDataviewConfig extends DataviewConfig<T = Type>
```

packages/layer-composer/src/generators/context/context-layers.ts
nerik: Gotcha... But the thing is, if I understand correctly, this value will never be picked up because you have a fallback <a href="https://github.com/GlobalFishingWatch/frontend/pull/194/files/682c55c93ee349ae58e40db99a141eda316868e8#diff-f5eb444a2dc0a126b3542d6aa5e1d2eecf166cc0d21f51b6d2b677c2a68cddf0R16">here</a> if config.color is not set

j8seangel: [Fixed](https://github.com/GlobalFishingWatch/frontend/pull/194/commits/4cd4400ddbd616f98c8ea1182291a9fd48c5b9be) thanks!

PR #198, opened 2020-11-10
Fishingmap/other context layers

j8seangel:
Refactor to share base config in context layers and include Tuna RFMOs and WPP NRI

Also:
- Handle error in use-tiles-state to remove the tile loading when a request fails
- Support custom order in style transformation to use if needed as for example in the labeler

### DIFF
--- a/packages/layer-composer/src/generators/context/context-layers.ts
+++ b/packages/layer-composer/src/generators/context/context-layers.ts
@@ -11,49 +11,71 @@ const settledBoundaries = [
   'Unilateral claim (undisputed)',
 ]
 
+const getDefaultContexInteraction = (): Partial<Layer> => {
+  return {
+    type: 'fill',
+    paint: {
+      'fill-color': 'transparent',
+      'fill-outline-color': 'transparent',
+    },
+    layout: {},
+    metadata: {
+      interactive: true,
+      group: Group.OutlinePolygons,
+    },
+  }
+}
+
+const getDefaultContexLine = (color: string): Partial<Layer> => {
+  return {
+    type: 'line',
+    paint: {
+      'line-color': color,
+    },
+    layout: {
+      'line-cap': 'round',
+      'line-join': 'round',
+    },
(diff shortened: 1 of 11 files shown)


packages/layer-composer/src/generators/context/context-layers.ts
```diff
@@ -11,49 +11,71 @@ const settledBoundaries = [
   'Unilateral claim (undisputed)',
 ]
 
+const getDefaultContexInteraction = (): Partial<Layer> => {
```
nerik, line 14: Nice!

nerik, line 14:
```suggestion
const getDefaultContextInteraction = (): Partial<Layer> => {
```

packages/layer-composer/src/generators/context/context-layers.ts
```diff
@@ -11,49 +11,71 @@ const settledBoundaries = [
...
+    },
+    layout: {},
+    metadata: {
+      interactive: true,
+      group: Group.OutlinePolygons,
+    },
+  }
+}
+
+const getDefaultContexLine = (color: string): Partial<Layer> => {
```
nerik, line 29:
```suggestion
const getDefaultContextLine = (color: string): Partial<Layer> => {
```

packages/layer-composer/src/generators/context/context-layers.ts
```diff
@@ -11,49 +11,71 @@ const settledBoundaries = [
...
+      'line-join': 'round',
+    },
+    metadata: {
+      interactive: false,
+      group: Group.OutlinePolygonsBackground,
+    },
+  }
+}
+
+const getDefaultContexLayersById = (id: string, color: string): Layer[] => {
```
nerik, line 46:
```suggestion
const getDefaultContextLayersById = (id: string, color: string): Layer[] => {
```

nerik (commented):
👍

PR #240, opened 2021-01-07
Polish 🇵🇱

nerik:
- [x] Always show all timebar settings options even when heatmap/vessels unavailable (but disabled)
- [x] Switch automatically to timebar track view when vessel is loaded for the 1st time
- [x] Fixed a bug where an empty timebar settings panel would show when using certain workspaces
- [x] Fixed a bug where a disabled radio button label would still be interactive
- [x] Show filters panels open by default when adding heatmap layer
- ~[ ] Let user activate bivariate when 3rd 4th layers etc are disabled~
- [x] prevent adding layers in combined/bivriate mode
- ~[ ] Some form of static graph to show *something* by default in timebar~

Later: Bivariate PR
- [ ] 5 colors bivariates
- [ ] add hover button _in between_ heatmap layers 
- [ ] move from `bivariate: true` to `bivariateStartIndex: 2`

### DIFF
--- a/.changeset/serious-insects-return.md
+++ b/.changeset/serious-insects-return.md
@@ -1,0 +1,6 @@
+---
+"@globalfishingwatch/ui-components": patch
+"@globalfishingwatchapp/fishing-map": patch
+---
+
+[WIP] Polish 🇵🇱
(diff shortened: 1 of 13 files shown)


j8seangel:
Nice round of polishing ✨ 
In the PR review I did some small changes:
- [x] Include missing translations
- [x] Move the `heatmapSublayersAddedIndex` from slice to `HeatmapSection` local state as the coming PR for permissions removes the `app.slice` simplifying the screenshotMode
- [x] Support bivariate mode not only when 2 layers but with 2 layers active

nerik:
@j8seangel Cool cool, thanks!

PR #245, opened 2021-01-13
aggregate per cell in interaction hook

nerik:
Aggregation is now done on a per-cell basis, when needed by the interaction hook - see https://github.com/GlobalFishingWatch/mapbox-gl-js/pull/108


j8seangel (approved):
👍

PR #246, opened 2021-01-13
Fishingmap/activity stacked area chart

nerik:
Screenshot 2021-01-18 at 15 59 28


applications/fishing-map/src/features/timebar/Timebar.tsx
```diff
@@ -39,9 +53,93 @@ const TimebarWrapper = () => {
...
+    let n = 0
+    n = performance.now()
+
+    // Getting features within viewport - it's somehow faster to use querySource with a crude viewport filter, than using queryRendered
+    const [boundsSW, boundsNE] = mapInstance.getBounds().toArray()
+    const allFeaturesWithStyle = mapInstance
+      .querySourceFeatures(currentTimeChunkId, {
+        sourceLayer: 'temporalgrid_interactive',
+      })
+      .filter((f) => {
```
j8seangel, line 88: Have you tried to use the [internal filter method](https://docs.mapbox.com/mapbox-gl-js/api/map/#map#querysourcefeatures) ? 

I guess it will have better performance as the number of features you have to manage into the component is smaller.

applications/fishing-map/src/features/timebar/Timebar.tsx
```diff
@@ -39,9 +53,93 @@ const TimebarWrapper = () => {
...
     [setBookmark]
   )
+  const mapInstance = useMapboxInstance()
+  const tilesLoading = useTilesState(mapInstance).loading
+  const currentTimeChunkId = useCurrentTimeChunkId()
+  const mapStyle = useMapStyle()
 
-  if (!start || !end) return null
+  const [stackedActivity, setStackedActivity] = useState<any>()
+  useEffect(() => {
```
j8seangel, line 62: I'd move what the effect does to a external method, for me it makes the component harder to read so what about creating something like `getStackedActivity()` and use it as: 
```
if (
   mapInstance ||
   mapStyle ||
   !tilesLoading ||
   timebarVisualisation === TimebarVisualisations.Heatmap
) {
  setStackedActivity(getStackedActivity(valuesByFrame))
}
```

j8seangel (changes requested):
It looks promising! ~however I couldn't manage to get it working as I never have rawValues, am I missing something?~
Please take a look on the comments and let me know your thoughts

Edit: it works like a charm! I fixed the conflicts to get it ready to be merged 
Also I just found a bug: the inactive layers are also present in the timebar, I think we should only render the layers which are visible:
Kapture 2021-01-22 at 13 17 16

applications/fishing-map/src/features/timebar/Timebar.tsx
j8seangel: as we discussed this works better, thanks for the explanation!

packages/layer-composer/src/generators/heatmap/util/fourwings.ts
```diff
@@ -0,0 +1,60 @@
+export const getCellValues = (rawValues: string) => {
+  // Raw values come as a single string (MVT limitation), turn into an array of ints first
+  const values: number[] = rawValues.split(',').map((v) => parseInt(v))
```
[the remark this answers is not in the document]
j8seangel: this is not happening anymore, I run `yarn install --force` it works now

PR #249, opened 2021-01-14
Fishing map/permissions

j8seangel:
- [x] Support guest users and introduce the concept of workspace public and private
- [x] Handle routing options errors
- [x] Simplify a lot the screenshotMode logic


applications/fishing-map/src/features/dataviews/dataviews.slice.ts
```diff
@@ -51,4 +51,6 @@ export const selectDataviewById = memoize((id: string) =>
   createSelector([(state: RootState) => state], (state) => selectById(state, id))
 )
 
+export const selectDataviewsStatus = (state: RootState) => state.datasets.status
```
nerik, line 54: I might be lacking context here, but shouldn't it be:
```suggestion
export const selectDataviewsStatus = (state: RootState) => state.dataviews.status
```

applications/fishing-map/src/features/map/MapScreenshot.tsx
```diff
@@ -15,21 +11,31 @@ declare global {
...
   px: number
   in: string
 }
 
-function MapScreenshot() {
-  const { screenshotMode, setScreenshotMode } = useScreenshotConnect()
-  const { setScreenshotLoading } = useScreenshotLoadingConnect()
-  const mapboxRef = useMapboxRef()
-  const documentFocus = useRef<boolean>(true)
+const isChrome = !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime)
```
nerik, line 19: I have some doubts about this:
- won't this yield false positives?
- do we correctly detect other Blink browsers such as Edge or Opera?
I would perhaps favor a more future proof solution ie using a 3rd party library (https://github.com/lancedikson/bowser for instance)

applications/fishing-map/src/features/workspace/Workspace.tsx
```diff
@@ -2,27 +2,108 @@ import React, { Fragment, useEffect } from 'react'
...
+          </Button>
+        ) : (
+          <Fragment>
+            <Button
+              href={`mailto:support@globalfishingwatch.org?subject=Requesting access for ${workspaceId} view`}
+            >
+              {t('errors.requestAccess', 'Request access') as string}
+            </Button>
+            <p className={styles.logged}>
+              {t('common.loggedAs', 'Logged as')} joseangel@globalfishingwatch.org
```
nerik, line 58: Is it still a requirement to be a Jose Angel to login to the map?

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -28,8 +28,24 @@ export const getDatasetsByDataview = (dataview: UrlDataviewInstance) =>
 
 export const selectWorkspace = (state: RootState) => state.workspace.data
 export const selectWorkspaceStatus = (state: RootState) => state.workspace.status
+export const selectWorkspaceError = (state: RootState) => state.workspace.error
 export const selectWorkspaceCustom = (state: RootState) => state.workspace.custom
 
+export const isWorkspacePublic = createSelector([selectWorkspace], (workspace) => {
+  return workspace?.public === true || workspace?.id.includes('-public')
```
nerik, line 35: Again, sorry if I'm missing context, but haven't we said `public` would be a prefix? This might lead to errors if for some reason a workspace has `public` in its name but not the the actual public prefix.
```suggestion
  return workspace?.public === true || workspace?.id.substr(0,6) === 'public'
```
(if it's a suffix, same idea)

applications/fishing-map/src/features/workspaces-list/WorkspacesList.tsx
```diff
@@ -14,18 +16,32 @@ import {
...
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [validCategory])
+
+  if (!validCategory) {
+    return (
+      <div className={styles.placeholder}>
+        <h2>{t('errors.pageNotFound', 'Page not found')}</h2>
+        <p>🙈</p>
```
nerik, line 38:
```suggestion
        <p>{['🙈','🙉','🙊'][Math.floor(Math.random()*3)]}</p>
```

nerik (changes requested):
I have a comment for you but I'll tell you on slack.

applications/fishing-map/src/features/dataviews/dataviews.slice.ts
j8seangel: ouch! 🤦

applications/fishing-map/src/features/workspace/workspace.selectors.ts
j8seangel: Finally it is going to be a suffix, but thanks for catching false positives! I'll fix it

applications/fishing-map/src/features/map/MapScreenshot.tsx
j8seangel: 👍

PR #251, opened 2021-01-20
Fishingmap/gear type

j8seangel:
Supporting gear type filter for the datasets which has the option, for now only `AIS`

https://user-images.githubusercontent.com/10500650/105179263-65860200-5b29-11eb-925d-38c33ad56ed7.mp4

A small of bug fixes I found:
- [x] Disable button on logout user to avoid sending the request twice
- [x] Not request the workspace again on search close


applications/fishing-map/public/locales/en/translations.json
```diff
@@ -36,6 +37,7 @@
     "filter_close": "Close filters",
     "filter_open": "Open filters",
     "flag_state_plural": "Flag States",
+    "gearType_plural": "Gear types",
```
fertandil87, line 40: the translator not support Pluralization?

applications/fishing-map/public/locales/source/translations.json
```diff
@@ -23,6 +23,7 @@
     "fishing": "Fishing"
```
fertandil87, line 23: source path is the fallback for the translations?

applications/fishing-map/src/features/workspace/heatmaps/heatmaps.utils.ts
```diff
@@ -1,5 +1,42 @@
...
+
+export const getCommonGearTypesInDataview = (dataview: UrlDataviewInstance) => {
+  const activeDatasets = dataview?.datasets?.filter((dataset) =>
+    dataview.config?.datasets.includes(dataset.id)
+  )
+  const geartypes = activeDatasets?.map((d) => d.schema?.geartype?.enum || [])
+  const commonGeartypes = geartypes
+    ? intersection(...geartypes).map((geartype) => ({
+        id: geartype,
+        label: capitalize(lowerCase(geartype)),
```
fertandil87, line 27: I prefer do the capitalize with css, but maybe is not possible with a label in the selector

applications/fishing-map/src/features/workspace/heatmaps/HeatmapFilters.tsx
```diff
@@ -18,15 +26,28 @@ function Filters({ dataview }: FiltersProps): React.ReactElement {
...
   const flags = useMemo(getFlags, [])
 
+  const datasetsWithoutgGeartype = getNotSupportedGearTypesDatasets(dataview)
   const onSelectSourceClick: MultiSelectOnChange = (source) => {
+    const datasets = [...(dataview.config?.datasets || []), source.id]
+    // We have to remove the geartype if it is not supported by the datasets selecion
+    const gearTypeOptions = getCommonGearTypesInDataview({ ...dataview, datasets })
+    const supportGearTypeFilter = gearTypeOptions.length > 0
+    const filters = dataview.config?.filters ? { ...dataview.config.filters } : {}
+    if (!supportGearTypeFilter && filters['geartype']) {
```
rdgfuentes, line 43: Thinking about a scenario where the user selects one source and for the gear types filters selects two: Drifting longlines and Set longlines. And then selects another source in addition than only supports Set longlines. Is the gear type filters selection updated?

applications/fishing-map/src/features/workspace/workspace.selectors.ts
```diff
@@ -197,16 +197,20 @@ export const selectDataviewInstancesResolved = createSelector(
...
       if (dataviewInstance.config?.type === Generators.Type.HeatmapAnimated) {
-        const dataviewInstanceWithUrlFilter = {
-          ...dataviewInstance,
-        }
-        if (dataviewInstance.config.filters?.length && dataviewInstanceWithUrlFilter.config) {
-          const flags = dataviewInstanceWithUrlFilter.config.filters
-          dataviewInstanceWithUrlFilter.config.filter = `flag IN (${flags
-            .map((f: string) => `'${f}'`)
-            .join(', ')})`
+        const { filters } = dataviewInstance.config
```
rdgfuentes, line 200: I think you could set the default value for `filters` here so that you don't need the if cause later.
```ts
const { filters = {} } = dataviewInstance.config
```

applications/fishing-map/src/features/workspace/heatmaps/HeatmapFilters.tsx
j8seangel: Good catch! I'll fix it, thanks for spotting it out

applications/fishing-map/src/features/workspace/workspace.selectors.ts
j8seangel: I like the approach but in this particular case doing it means I'll need later to check if it is an empty array to avoid doing
```
Object.keys(filters)...
```
so does it really change something ?

applications/fishing-map/public/locales/source/translations.json
j8seangel: Yes it should

applications/fishing-map/src/features/workspace/heatmaps/heatmaps.utils.ts
j8seangel: as the `Select` is a reusable component... do you think does it worth to generalise this behaviour?

applications/fishing-map/src/features/workspace/heatmaps/HeatmapFilters.tsx
j8seangel: Fixed in https://github.com/GlobalFishingWatch/frontend/commit/24b604f3acb9e877c986c2f7f680d737ee5b016f
I had to mock VMS Indonesia dataset to include `drifting_longlines` and `tuna_purse_seines` but now the selection is updated with the options both datasets shares, an example as it is nos possible to test without the mock:

https://user-images.githubusercontent.com/10500650/105324930-52783e00-5bcc-11eb-83fd-3fb087281138.mp4

applications/fishing-map/src/features/workspace/workspace.selectors.ts
rdgfuentes: It's just a minor improvement, it does not change the behavior. Note that if `filters` receives its default value `{}` (empty object) then it's not necessary to check anything since `Object.keys(filters)` will be an empty array and thus `sqlFilters` too.

PR #255, opened 2021-01-21
Fishingmap/search permissions

j8seangel:
Search only in datasets where the user has permissions, see for more login details: https://globalfishingwatch.atlassian.net/browse/MAP-219

**TLTR**: filter datasets where the user can search => [see code](https://github.com/GlobalFishingWatch/frontend/compare/fishingmap/search-permissions?expand=1#diff-0fb40e3c620de822cefcb24ad9970369ff573312c02d2b8e577e08323a163324R6)

But this PR also:
- [x] Fixes when refreshing in search page
- [x] Include states when no search datasets available.
image

- TODO: move the `auth-middleware` package from GitHub to NPM and hopefully include typings


applications/fishing-map/src/features/app/App.tsx
```diff
@@ -44,6 +53,18 @@ function App(): React.ReactElement {
...
+    if (userLogged) {
+      if (
+        (locationType === HOME && workspaceStatus !== AsyncReducerStatus.Finished) ||
+        (locationType !== WORKSPACE && currentWorkspaceId !== workspaceId)
+      ) {
+        dispatch(fetchWorkspaceThunk(workspaceId as string))
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [userLogged, workspaceId])
```
rdgfuentes, line 66: Is possible that `locationType` or `currentWorkspaceId` changes and `workspaceId` stays unmodified? Not sure if it's a valid scenario or not, if so then maybe it would convenient to add them to the array so the `useEffect` hook is executed when they change too.

applications/fishing-map/src/features/search/Search.tsx
```diff
@@ -41,33 +45,37 @@ function Search() {
...
+          fetchVesselSearchThunk({
+            query: debouncedQuery,
+            filters: searchFilters,
+            datasets: sources,
+            offset,
+          })
+        )
+      }
     },
     [debouncedQuery, dispatch, searchDatasets, searchFilters]
```
rdgfuentes, line 80: Should we add `offset` to this variables list too?

applications/fishing-map/src/features/app/App.tsx
j8seangel: It wasn't possible but as seems to be confusing I've made it more clear, take a look at: https://github.com/GlobalFishingWatch/frontend/commit/cf267d75b9c532e647e96a863d61ab354b2b99e2
Also I got a bug when going to the home from a workspace without the loading finished caused not triggering the default view fetch, it was also fixed and added a validation to avoid multiple fetch.

rdgfuentes: Looks great! 👍

PR #257, opened 2021-01-21
added requestType to the fetch api

fertandil87:
This improvement is related to upload a formData body from the labeler 

https://github.com/GlobalFishingWatch/track-labeler/pull/29

### DIFF
--- a/packages/api-client/src/api-client.ts
+++ b/packages/api-client/src/api-client.ts
@@ -1,5 +1,5 @@
 import { saveAs } from 'file-saver'
-import { UserData, ResourceResponseType } from '@globalfishingwatch/api-types'
+import { UserData, ResourceResponseType, ResourceRequestType } from '@globalfishingwatch/api-types'
 import { isUrlAbsolute } from './utils/url'
 
 const API_GATEWAY =
@@ -34,6 +34,7 @@ interface LoginParams {
 export type FetchOptions<T = BodyInit> = Partial<RequestInit> & {
   method?: 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE'
   responseType?: ResourceResponseType
+  requestType?: ResourceRequestType
   dataset?: boolean
   body?: T
   local?: boolean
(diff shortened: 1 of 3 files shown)


packages/api-client/src/api-client.ts
```diff
@@ -239,10 +241,10 @@ export class GFWAPI {
         const data = await fetch(fetchUrl, {
           method,
           signal,
-          ...(body && { body: JSON.stringify(body) }),
+          ...(requestType === 'json' ? body && { body: JSON.stringify(body) } : body && { body }),
```
j8seangel, line 244: The objective of the spread operator in the body is to include the property only when the body is defined, to understand how it works take a look at: https://medium.com/@mikeh91/conditionally-adding-keys-to-javascript-objects-using-spread-operators-and-short-circuit-evaluation-acf157488ede
With this change, at least for me, it harder to understand what is happening so my suggestion to simplify it is:
```suggestion
          ...body && { body: requestType === 'json' ? JSON.stringify(body): body },
```

packages/api-client/src/api-client.ts
```diff
@@ -239,10 +241,10 @@ export class GFWAPI {
...
           headers: {
             ...headers,
-            ...(responseType === 'json' && { 'Content-Type': 'application/json' }),
+            ...(requestType === 'json' && { 'Content-Type': 'application/json' }),
```
rdgfuentes, line 247: So the `Content-Type` is now dependent of the `requestType` only, is that correct? By looking at the function signature it can receive both `requestType` and `responseType` and they could not match. What would be the expected header to send in the request when `responseType = 'json'` and `requestType = 'formData'`? Is this a valid scenario?

fertandil87: @rdgfuentes the response type is for this:
.then((res) => {
            switch (responseType) {
              case 'default':
                return res
              case 'json':
                return parseJSON(res)
              case 'blob':
                return res.blob()
              case 'text':
                return res.text()
              case 'arrayBuffer':
                return res.arrayBuffer()
              case 'vessel': {
                try {
                  return import('@globalfishingwatch/pbf/decoders/vessels.js').then(

packages/api-client/src/api-client.ts
j8seangel: Thanks for the change, LGTM!

PR #272, opened 2021-01-29
Vessel history/vh 57

fertandil87:
THIS PR NEED A REBASE, DON'T SEE IT

https://globalfishingwatch.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=VH&modal=detail&selectedIssue=VH-57
image
image


applications/vessel-history/src/features/home/Home.tsx
```diff
@@ -63,29 +65,33 @@ const Home: React.FC<LoaderProps> = (): React.ReactElement => {
       })
       .catch((error) => {
         setSearching(false)
+        return vessels
         /*this.setState({
         tip: this.tips.connectionError,
       })*/
       })
+    console.log(newVessels)
```
rdgfuentes, line 73: Remove console.log

applications/vessel-history/src/features/home/Home.tsx
```diff
@@ -109,19 +115,15 @@ const Home: React.FC<LoaderProps> = (): React.ReactElement => {
...
         )}
         {query && (
           <div>
-            <h2>VESSELS FOUND</h2>
+            {searching && <Spinner className={styles.loader}></Spinner>}
             <div className={styles.offlineVessels}>
-              {vessels.map((vessel) => (
-                <VesselListItem vessel={vessel} />
+              {vessels.map((vessel, index) => (
+                <VesselListItem key={index} vessel={vessel} />
```
rdgfuentes, line 126: 👍

applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
```diff
@@ -1,14 +1,35 @@
...
+import IconButton from '@globalfishingwatch/ui-components/src/icon-button/IconButton'
+import { Extra, Vessel } from 'types'
+import { getFlagById } from 'utils/flags'
 import styles from './VesselListItem.module.css'
 import '@globalfishingwatch/ui-components/dist/base.css'
-
 interface ListItemProps {
   saved?: boolean
-  vessel: any
+  vessel: Vessel
```
rdgfuentes, line 10: 👍

applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
```diff
@@ -19,31 +40,43 @@ const VesselListItem: React.FC<ListItemProps> = (props): React.ReactElement => {
...
+          </div>
+        )}
         <div>
           <label>TRANSMISSIONS</label>
-          from 2015-07-21 to 2020/04/13
-        </div>
-        <div>
-          <label>IMO</label>
-          12345
+          from {vessel.firstTransmissionDate} to {vessel.lastTransmissionDate}
```
rdgfuentes, line 79: Should dates be formatted?

applications/vessel-history/src/utils/flags.ts
```diff
@@ -0,0 +1,11 @@
+import flags from 'data/flags'
+
+type Flag = { id: string; label: string }
+export const getFlagById = (id: string): Flag | undefined => {
+  const flag = flags.find((f) => f.id === id)
```
rdgfuentes, line 5: It would be more performant to have the flags stored in a dictionary instead of an array. As of now we are only accessing them by Id, thus it would be much faster than iterating the flags array looking for the given id. Makes sense?

rdgfuentes (changes requested):
See my comments below

applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
fertandil87: no, is the same format of the design, and not have timezone offset

applications/vessel-history/src/utils/flags.ts
fertandil87: @rdgfuentes this comes from another project of the monorepo, my idea was to join this in a package then, not update it, I don't know why they implemented in this way. I know that have a list of flags (and also duplicated in two or more projects) is not the best solution, but I think the first improvement need to move this functionality (from all the apps) to a centralized place, and then see the implementation as you said, what do you think @j8seangel ?

PR #273, opened 2021-01-29
Vessel history/VH-37 Splash & Login (no authorization)

rdgfuentes:
Jira: https://globalfishingwatch.atlassian.net/browse/VH-37
- [x]  Display splash screen when page is loading
image

- [x] User can login and logout to the app


fertandil87 (approved):

fertandil87:
@rdgfuentes this branch need a rebase

package.json
```diff
@@ -50,21 +51,23 @@
...
-    "@typescript-eslint/eslint-plugin": "^4.6.1",
+    "@typescript-eslint/eslint-plugin": "^4.14.1",
     "@typescript-eslint/parser": "^4.6.1",
     "autoprefixer": "^9.8.0",
     "babel-eslint": "^10.1.0",
     "cross-env": "^7.0.2",
     "cssnano": "^4.1.10",
     "eslint": "7.12.1",
-    "eslint-config-prettier": "6.15.0",
+    "eslint-config-prettier": "^7.2.0",
```
j8seangel, line 61: Thanks for update them

package.json
```diff
@@ -25,6 +25,7 @@
     "build": "lerna run build --scope @globalfishingwatch/*",
     "test": "lerna run test --scope @globalfishingwatch/*",
     "test:prepare": "lerna run prepare --scope @globalfishingwatch/*",
+    "test:vessel-history": "cross-env lerna run test --scope @globalfishingwatchapp/vessel-history",
```
j8seangel, line 28: I don't think so... but would there be a way to pass the test:`:[command]` as a variable ? so we avoid duplicating the same command for every project, anyway it looks good

applications/vessel-history/src/components/Splash.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { memo } from 'react'
```
j8seangel, line 1: I don't think we need to always render a Splash component with a certain timeout, why not include loading spinners on the UI when needed ?

applications/vessel-history/src/components/Splash.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import Splash from './Splash'
+
+describe('Splash', () => {
```
j8seangel, line 5: Is this just testing that the component doesn't break? 
Just to understand how it works, if props are need, does it complain?

j8seangel (changes requested):
See comment about <Splash /> because I don't think it is a good practice, but lets talk and find an approach that works for both

applications/vessel-history/src/components/Splash.test.tsx
rdgfuentes: Right, this component does not receive any props. It's just testing that the component rendered matches the snapshot that was generated and saved for it. When it fails we'll have to decide whether we have to update the snapshot becase the change is expected or fix the component to continue rendering what we expected (snapshot)

applications/vessel-history/src/components/Splash.tsx
rdgfuentes: You are right we can just remove the timeout, we only introduced it so that the splash screen is noticeable in faster internet connection speeds.

package.json
rdgfuentes: Not that I know of...

applications/vessel-history/src/App.tsx
```diff
@@ -18,10 +19,13 @@ function App() {
     const location = window.location.origin + BASE_URL
     window.location.href = GFWAPI.getLoginUrl(location)
   }
-  //const search = useSelector(getLocationSearch)
 
+  //const search = useSelector(getLocationSearch)
```
enriquetuya, line 23: If code is not used, delete it.

applications/vessel-history/src/components/Splash.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Logo from '@globalfishingwatch/ui-components/dist/logo'
+import vesselHistoryLogo from '../assets/images/splash-screen-image@2x.png'
+import tmtLogo from '../assets/images/tmt_logo_final_full_colour@2x.png'
+import styles from './Splash.module.css'
+
+function Splash() {
+  return (
+    <div className={styles.container}>
+      <div className={styles.splash}>
+        <img src={vesselHistoryLogo} className={styles.vhLogo} alt="Vessel History" />
```
enriquetuya, line 12: The name `Vessel History` is not set we need to remember all places where we are stating the name in case we need to change it.

enriquetuya (commented):
@rdgfuentes @fertandil87 just two little comments.

applications/vessel-history/src/components/Splash.tsx
rdgfuentes: Once I18N is in place later in this sprint (https://globalfishingwatch.atlassian.net/browse/VH-54) we'll be able to store all strings in the `translations.json` files, makes sense?

applications/vessel-history/src/components/Splash.tsx
fertandil87: @j8seangel @rdgfuentes I think that the ticket require that the splash show for a moment (maybe to see the icon of sponsor), or I am wrong?

applications/vessel-history/src/components/Splash.tsx
fertandil87: makes sense for me

applications/vessel-history/src/components/Splash.test.tsx
j8seangel: 👌

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
+{
```
j8seangel, line 1: I'd like to know how to use it, could you do a short demo on the frontend meeting we have on Thursday ?
I think it should be here:
image
but I don't know all the features it has

j8seangel (approved):
Code looks good to me!
I had some warnings as there are unused variables but we could fix them progressively 
image

.vscode/launch.json
rdgfuentes: Sure, sounds good

PR #297, opened 2021-02-05
VH-39 Integrate with TMT Search API

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VH-39
- Use TMT API endpoint for searches
- Updated Vessel model


fertandil87 (approved):

applications/vessel-history/src/types/index.ts
```diff
@@ -21,40 +21,35 @@ export type MapCoordinates = {
...
-  type: string
-  ssvid: string
-  name: string
-  imo: string
-  flags: Flag[]
-  authorizations: any[]
-  mmsi: Mmsi[]
-  callsign: Callsign[]
-  extra: Extra[]
+  callsign: string
```
j8seangel, line 42: ~isn't this a breaking change with the old API~ ? I just realised not depends on the API anymore but on the dataset so I guess we'll need to support generics here

j8seangel (commented):
Apart of the API definition type LGTM

PR #298, opened 2021-02-05
VH-34 Move Vessel History to monorepo

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VH-34
 First version of vessel history to include in the monorepo.


applications/vessel-history/package.json
```diff
@@ -0,0 +1,144 @@
...
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  "jest": {
```
j8seangel, line 133: Is there a way to centralice the configuration ? The same that we have with tsconfig and rollup config

applications/vessel-history/public/index.html
```diff
@@ -0,0 +1,114 @@
...
+    media="(device-width: 1024px) and (device-height: 1366px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
+    href="%PUBLIC_URL%/icons/apple-touch-startup-image-2732x2048.png">
+  <link rel="apple-touch-startup-image"
+    media="(device-width: 810px) and (device-height: 1080px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
+    href="%PUBLIC_URL%/icons/apple-touch-startup-image-2160x1620.png">
+  <meta name="msapplication-TileColor" content="#fff">
+  <meta name="msapplication-TileImage" content="%PUBLIC_URL%/icons/mstile-144x144.png">
+  <meta name="msapplication-config" content="%PUBLIC_URL%/icons/browserconfig.xml">
+
+  <title>Global Fishing Watch | Sandbox</title>
```
j8seangel, line 100: There are some typos on the title as it boilerplate comes from the Sandbox

applications/vessel-history/src/App.tsx
```diff
@@ -0,0 +1,44 @@
...
+import useGFWLogin from '@globalfishingwatch/react-hooks/dist/use-login'
+import Home from 'features/home/Home'
+import { getLocationType } from 'routes/routes.selectors'
+import { PROFILE } from 'routes/routes'
+import { BASE_URL, SPLASH_TIMEOUT } from 'data/constants'
+import Profile from 'features/profile/Profile'
+import Splash from 'components/Splash'
+import './App.css'
+
+GFWAPI.setConfig({ dataset: process.env.REACT_APP_DATASET || 'carriers:v20200820' })
```
j8seangel, line 13: Legacy GFWAPI stuff... we don't use a fixed version of the dataset anymore so we could drop it off

applications/vessel-history/src/components/Loader.js
```diff
@@ -0,0 +1,10 @@
+import React, { Component } from 'react'
+import styles from './Loader.module.css'
```
j8seangel, line 2: There is a <Spinner /> component in ui-components, I'd use it

applications/vessel-history/src/features/profile/components/Event.js
```diff
@@ -0,0 +1,92 @@
+import React, { Component } from 'react'
+// import { Link } from 'react-router-dom'
+import cx from 'classnames'
+import dayjs from 'dayjs'
```
j8seangel, line 4: Let's talk about date format libraries in next frontend meeting to align on using the same across projects

applications/vessel-history/src/features/profile/components/MapWrapper.js
```diff
@@ -0,0 +1,64 @@
+import React, { useState, useEffect } from 'react'
+import PropTypes from 'prop-types'
+import geobuf from 'geobuf'
+import Pbf from 'pbf'
+import GFWAPI from '@globalfishingwatch/api-client'
+import { InteractiveMap } from '@globalfishingwatch/react-map-gl'
```
j8seangel, line 6: 👍

applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react'
+import Logo from '@globalfishingwatch/ui-components/src/logo'
+import { IconButton } from '@globalfishingwatch/ui-components'
+import { Extra, Vessel } from 'types'
+import { getFlagById } from 'utils/flags'
+import styles from './VesselListItem.module.css'
+import '@globalfishingwatch/ui-components/dist/base.css'
```
j8seangel, line 7: This is only needed once.

j8seangel (changes requested):
Good amount of work here! Please take a look on the comments

applications/vessel-history/src/features/profile/components/Event.js
nerik: As a base to discuss on the next meeting, here's our experience with date/time libraries so far:
- moment.js is too big, unless stripping some of its codebase, and its own authors mention it as deprecated
- date-fns is lightweight but syntax is awkward
- dayjs has a nice, lean, moment-like API, but it's **UTC time/timezones capabilities**, which are crucial in GFW, are not good/reliable enough (or were when we checked)
- we settled for now on <a href="https://moment.github.io/luxon/">Luxon</a>, from moment creators:
  - it has an clear and explicit API - no magic formats like 'LLLL', more like DateTime.DATE_MED
  - it handles UTC time well
  - it is lightweight 
  - it provides 18n capabilities by relying on native browser i18n capabilities
  
Let's discuss that, also taking the opportunity to remind us that it's important to **always work with UTC times** in GFW projects. We had a lot of errors caused by not paying attention to that (although in Argentina if you use local time by mistake the problem should more obvious, but for us we are just 1hr away from UTC time)  

Example: 
https://github.com/GlobalFishingWatch/frontend/blob/f5519e6b8eb57201efd792c4fe4af674f478c67b/packages/layer-composer/src/generators/heatmap/util/time-chunks.ts#L40

applications/vessel-history/src/App.tsx
rdgfuentes: This is removed in #297

applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
rdgfuentes: 👍

applications/vessel-history/package.json
rdgfuentes: Yes, there's an approach that lets define a jest.config.base.js in the root folder of the monorepo and then include it in the jest.config.js of each project. I tried to setup it that way but it did not work. We could give it another try later or talk about it.

applications/vessel-history/src/features/profile/components/Event.js
rdgfuentes: Makes sense, let's use luxon from now on, this code was not revisited yet so still needs to be cleaned.

package.json
```diff
@@ -102,6 +102,9 @@
     "react": "^7.0.1",
     "react-dom": "^17.0.1"
   },
+  "resolutions": {
```
j8seangel, line 105: is this going to force every version of babel-loader to use 8.1.0? or what does it do exactly?

fertandil87 (approved):

package.json
rdgfuentes: Right, we could set it to  "^8.1.0" if you prefer. Or should we use the `SKIP_PREFLIGHT_CHECK=true` env variable when there are version conflicts in the packages that use our dependencies?

j8seangel: let's use this fix for now as seems to be more elegant than the `SKIP_PREFLIGHT_CHECK` for now.
I've read we should change a bit how we manage the dependencies in the Monorepo and each package.json responsibility but this will be addressed later on. Thanks for the clarification!

rdgfuentes:
Thanks @fertandil87 you did most of the work here 👏

PR #300, opened 2021-02-09
Fishingmap/ocean areas miniglobe

nerik:
Adds textual information about the current area when hovering over mini globe.
Additionally, fixes some issues/adds details to the underlying `ocean-areas` library.

https://user-images.githubusercontent.com/1583415/107380157-ea8f8600-6aed-11eb-8ac6-45bf4c07e7fa.mov


j8seangel (commented):
I see the data file creases the size a lot:
image 
is there a way to minimise it? maybe storing as a topojson? because having 500kb loaded in every bundle is a lot for me on a feature than a user might not use ever.

Also I'm missing the new method documentation on the README.md, could you please add it?

PD: you missed the changeset, otherwise the deploy will fail

nerik:
@j8seangel We managed to divide the size of the file by ~2, it's now 307k.
Also note that prettify is adding a lot of whitespace (170k without line breaks), which is irrelevant for the final bundle. Gzipped the file ends up being 45k.

satellitestudiodesign:
Cool, cool cool cool.

PR #301, opened 2021-02-09
Fishingmap/datasets advance upload

j8seangel:
Finish moving amathea upload dataset features by supporting dataset with color ranges in fishing-map
Uploading as:
image
Render as:
image

### DIFF
--- a/applications/fishing-map/src/features/datasets/NewDataset.tsx
+++ b/applications/fishing-map/src/features/datasets/NewDataset.tsx
@@ -8,8 +8,8 @@ import type { FeatureCollectionWithFilename } from 'shpjs'
 import InputText from '@globalfishingwatch/ui-components/dist/input-text'
 import Modal from '@globalfishingwatch/ui-components/dist/modal'
 import Button from '@globalfishingwatch/ui-components/dist/button'
-import Select, { SelectOption } from '@globalfishingwatch/ui-components/dist/select'
-import { DatasetTypes } from '@globalfishingwatch/api-types'
+import Select from '@globalfishingwatch/ui-components/dist/select'
+import { DatasetConfiguration, DatasetTypes } from '@globalfishingwatch/api-types'
 import { ReactComponent as FilesIcon } from 'assets/icons/files-supported.svg'
 import { capitalize } from 'utils/shared'
 import { SUPPORT_EMAIL } from 'data/config'
@@ -26,6 +26,25 @@ const extractPropertiesFromGeojson = (geojson: GeoJSON.FeatureCollection) => {
     }, {})
   )
   return uniqueProperties
+}
+
+const getPropertyRangeValuesFromGeojson = (
+  geojson: GeoJSON.FeatureCollection,
+  property: string
+) => {
+  if (!geojson?.features) return
+  const propertyRange = geojson.features.reduce(
+    function (acc, { properties }) {
+      const value = properties?.[property]
+      if (!value) return acc
+      return {
+        min: value < acc.min ? value : acc.min,
+        max: value > acc.max ? value : acc.max,
+      }
+    },
+    { min: 0, max: 0 }
+  )
+  return propertyRange
 }
 
 interface DatasetConfigProps {
@@ -103,13 +122,24 @@ const DatasetConfig: React.FC<DatasetConfigProps> = (props) => {
           className={styles.selectShort}
           containerClassName={styles.selectContainer}
           options={colorByOptions}
-          selectedOption={metadata.configuration?.colorBy}
+          selectedOption={colorByOptions.find(
+            ({ id }) => id === metadata.configuration?.propertyToInclude
+          )}
           onSelect={(selected) => {
             // TODO: pre-populate min and max values depending on selection
-            onDatasetFieldChange({ colorBy: selected })
+            onDatasetFieldChange({ propertyToInclude: selected.id })
+            const propertyRange = getPropertyRangeValuesFromGeojson(fileData, selected.id)
+            if (propertyRange) {
+              onDatasetFieldChange({
+                propertyToIncludeRange: {
+                  min: propertyRange.min,
+                  max: propertyRange.max,
+                },
+              })
+            }
           }}
           onRemove={() => {
-            onDatasetFieldChange({ colorBy: undefined })
+            onDatasetFieldChange({ propertyToInclude: undefined })
           }}
           direction="top"
         />
@@ -117,14 +147,14 @@ const DatasetConfig: React.FC<DatasetConfigProps> = (props) => {
           inputSize="small"
           type="number"
           step="0.1"
-          value={metadata.configuration?.colorValues?.min}
+          value={metadata.configuration?.propertyToIncludeRange?.min}
           placeholder={t('common.min', 'Min')}
           className={styles.shortInput}
           onChange={(e) =>
             onDatasetFieldChange({
-              colorValues: {
+              propertyToIncludeRange: {
                 min: parseFloat(e.target.value),
-                max: metadata.configuration?.colorValues?.max,
+                max: metadata.configuration?.propertyToIncludeRange?.max,
               },
             })
           }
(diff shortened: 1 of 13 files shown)


applications/fishing-map/src/features/map/Map.tsx
```diff
@@ -144,6 +148,7 @@ const MapWrapper = (): React.ReactElement | null => {
   // useLayerComposer is a convenience hook to easily generate a Mapbox GL style (see https://docs.mapbox.com/mapbox-gl-js/style-spec/) from
   // the generatorsConfig (ie the map "layers") and the global configuration
   const { style } = useLayerComposer(generatorsConfig, globalConfig)
+  console.log(style)
```
j8seangel, line 151: 🙈 oops, already removed in dev

rdgfuentes, line 151: Remove

packages/api-types/src/datasets.ts
```diff
@@ -41,11 +41,16 @@ export enum DatasetStatus {
   Error = 'error',
 }
 
+export type DatasetCustomTypes = 'points' | 'lines' | 'geometries'
 export interface DatasetConfiguration {
   index?: string
   filePath?: string
-  propertyToInclude?: string
   srid?: number
+  file?: string
```
rdgfuentes, line 49: Defining [optional properties for our interfaces is not a good habit in typescript](https://startup-cto.net/10-bad-typescript-habits-to-break-this-year/#6optionalproperties). In general this is a sign that the interface could still be extended in several data types that are more explicit and will provide us type checking at compile-time instead of run-time.
In this specific case maybe we should define `PointsDatasetConfiguration`, `LinesDatasetConfiguration` and `GeometriesDatasetConfiguration`, each one containing only the properties that are needed.

applications/fishing-map/src/features/datasets/NewDataset.tsx
```diff
@@ -80,36 +113,69 @@ const DatasetConfig: React.FC<DatasetConfigProps> = (props) => {
...
+          type="number"
+          step="0.1"
+          value={metadata.configuration?.propertyToIncludeRange?.min}
           placeholder={t('common.min', 'Min')}
           className={styles.shortInput}
-          onChange={(e) => console.log(e.target.value)}
+          onChange={(e) =>
+            onDatasetFieldChange({
+              propertyToIncludeRange: {
+                min: parseFloat(e.target.value),
```
rdgfuentes, line 156: Where do we validate that the min is not greater than the max ?

applications/fishing-map/src/features/map/map.selectors.ts
```diff
@@ -146,6 +149,20 @@ export const getWorkspaceGeneratorsConfig = createSelector(
           if (dataset?.source) {
             generator.attribution = dataset.source
           }
+          if (dataview.config?.type === Generators.Type.UserContext) {
+            if (
+              dataset.configuration?.propertyToInclude &&
```
rdgfuentes, line 154: This condition is in other places too, what is it used for? could we add a method to the interface that defines a contract for this condition and then implement it when the dataset configuration is created, makes sense?

applications/fishing-map/src/features/datasets/datasets.hook.ts
```diff
@@ -20,16 +26,30 @@ import {
...
 
   const addNewDatasetToWorkspace = useCallback(
-    (dataset) => {
-      const dataviewInstance = getContextDataviewInstance(dataset.id, usedColors)
-      upsertDataviewInstance(dataviewInstance)
+    (dataset: Dataset) => {
+      let dataviewInstance
+      const datasetType =
+        dataset.configuration?.propertyToInclude && dataset.configuration?.propertyToIncludeRange
+          ? 'enviromental'
```
rdgfuentes, line 38: There is an [enum defined for `DatasetTypes`](https://github.com/GlobalFishingWatch/frontend/blob/a9e66b1c44f33892904339582479772e472a0a3a/packages/api-types/src/datasets.ts#L28), should it be used here?

applications/fishing-map/src/features/datasets/NewDataset.tsx
```diff
@@ -159,16 +225,11 @@ const DatasetFile: React.FC<DatasetFileProps> = ({ onFileLoaded, className = ''
...
 export type DatasetMetadata = {
   name: string
   description?: string
   type: DatasetTypes.Context
-  configuration?: {
-    file?: string
-    type?: DatasetCustomTypes
-    format?: 'geojson'
-  }
+  configuration?: DatasetConfiguration
```
rdgfuentes, line 232: Thanks for using an interface here. Is this field still optional?

applications/fishing-map/src/features/datasets/NewDataset.tsx
```diff
@@ -18,17 +18,50 @@ import { readBlobAs } from 'utils/files'
...
+  if (!geojson?.features) return []
+  const uniqueProperties = Object.keys(
+    geojson.features.reduce(function (acc, { properties }) {
+      return { ...acc, ...properties }
+    }, {})
+  )
+  return uniqueProperties
+}
+
+const getPropertyRangeValuesFromGeojson = (
```
rdgfuentes, line 31: 👍

applications/fishing-map/src/features/datasets/NewDataset.tsx
```diff
@@ -18,17 +18,50 @@ import { readBlobAs } from 'utils/files'
 import { useDatasetsAPI, useDatasetModalConnect, useNewDatasetConnect } from './datasets.hook'
 import styles from './NewDataset.module.css'
 
+const extractPropertiesFromGeojson = (geojson: GeoJSON.FeatureCollection) => {
```
rdgfuentes, line 21: 👍

rdgfuentes (commented):
@j8seangel Good work! Take a look at my comments below

applications/fishing-map/src/features/datasets/NewDataset.tsx
```diff
@@ -18,17 +18,50 @@ import { readBlobAs } from 'utils/files'
...
+      }
+    },
+    { min: 0, max: 0 }
+  )
+  return propertyRange
+}
+
 interface DatasetConfigProps {
   className?: string
   onDatasetFieldChange: (field: any) => void
```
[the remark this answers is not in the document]
j8seangel: Yes, I forgot to replace it here, already using:
```
DatasetMetadata | DatasetConfiguration
```

applications/fishing-map/src/features/datasets/NewDataset.tsx
j8seangel: Done!
image

applications/fishing-map/src/features/datasets/NewDataset.tsx
j8seangel: Yes, a very basic dataset doesn't need configuration.

applications/fishing-map/src/features/datasets/datasets.hook.ts
j8seangel: This is just a workaround in the frontend to distinguish between environmental layer and context areas datasets and it is not related with the datasetType. So to make it clear I've renamed it to `layerType`

applications/fishing-map/src/features/map/map.selectors.ts
j8seangel: The same for the workaround before, as it won't be needed when we have the different layer types feature it will be removed

j8seangel: Next time I'll leave some comments on the code.

packages/api-types/src/datasets.ts
j8seangel: Thanks for sharing! I'll start by the basic of having `DatasetConfiguration` and `EnviromentalDatasetConfiguration` and then lets discuss how to split them by other types.

I also created a new `type AnyDatasetConfiguration = DatasetConfiguration | EnvDatasetConfiguration` to update the dataset configuration type to use:
```js
configuration: AnyDatasetConfiguration | null
```
does it makes sense?

rdgfuentes: Thanks @j8seangel! Great approach 💯 !

PR #303, opened 2021-02-09
MAP-233 check duplicated uniqueId generated by aggregate-tile

rdgfuentes:
DO NOT MERGE Just to share some tests I did to search for duplicate "uniqueId"


nerik:
@rdgfuentes @j8seangel 
Couple of things:
- The test you added was not correct in that it couldn't possibly yield different results: `generateUniqueId` is deterministic, so if you do it 20 times you should get the same result (while the id is unique, it's not random)
- I moved `generateUniqueId` to an util so that it can be tested separately
- The issue was that `generateUniqueId` could generate string ids with leading 0s, which then were `parseInt`ed so the 0s would get dropped.
- A separate "issue" was that if zoomed out enough, you will highlight cells that belong to the same area of the world (case where several copies of the world are on screen). Discussed with @satellitestudiodesign and we consider it's a bit weird, but not incorrect, so good to go.

rdgfuentes:
Good catch @nerik . On item (4) I agree it's weird but not incorrect.

PR #311, opened 2021-02-17
Fishingmap/enviromental datasets upload

j8seangel:
Upload datasets by context and environment categories:
- [x] Refactor code to make context functionalities reusable
- [x] Use refactored code for both categories
- [x] Support edit values for environment datasets  

Included some documentation for the used dataviews, [take a look 📝  👀 ](https://github.com/GlobalFishingWatch/frontend/blob/79215bb64a133f51cf31588c01bf72d4e3ee5255/applications/fishing-map/src/data/dataviews.md)


packages/api-types/src/dataviews.ts
```diff
@@ -1,8 +1,8 @@
 import { ApiAppName, Dataset } from '.'
 
-export interface DataviewConfig<T = any> {
+export interface DataviewConfig<Type = any> {
   // TODO use any property from layer-composer here?
-  type?: T
+  type?: Type
```
rdgfuentes, line 5: thanks, now it's more clear.

j8seangel: no, thanks to you! I got it from the typescript errors to avoid link you sent

PR #327, opened 2021-02-19
Added opacity option to track layer

fertandil87:
I am adding a new property for the track layer  configuration to support set the opacity of the line


rdgfuentes (approved):
LGTM

PR #347, opened 2021-03-01
MAP-122 Move to shared components

rdgfuentes:
- [x] Moved ReportLayerPanel to shared
- [x] Created DatasetSchemaField component
- [x] Fixed report to be not available when no activity layer is not visible


j8seangel:
Thanks for this! I've merged [into my current branch](https://github.com/GlobalFishingWatch/frontend/commit/6dc89901387226af182c7b10a325f9e57b449513) also moving the sources and the flag fields to a new component and looks much simpler now in HeatmapLayerPanel and AnalysisLayerPanel without duplicated logic:

image

PR #348, opened 2021-03-03
Fishingmap/map 184 flying reports

j8seangel:
A lot of things happening but we have this `very unstable WIP` graph 📈 💯 

Cool things already working:
- Reload the graph on precision changes, so when going from 10days to day will update the values to better precision
- Reloading url works!
  - [x] fit to zoom 0 at the beginning to ensure geometry of the area is loaded
  - [x] fit to area bounds once loaded
- Identify overlapping cells to estimate the error

image

To improve:
~- [ ] Use only the percentage of the fields overlapping the area to reduce the error instead of the entire cell~
~- [ ] Race condiciones on source loaded~
Fixed on https://github.com/GlobalFishingWatch/frontend/pull/356

Out of scope things I need to do:
- Use [recoil](https://recoiljs.org) to persist the viewport state in `useViewport` hook across instances
- Use [workerize-loader](https://github.com/developit/workerize-loader) to use a worker in the heavy process of the filterByPolygon


applications/fishing-map/src/features/analysis/Analysis.tsx
```diff
@@ -0,0 +1,243 @@
...
+        cleanFeatureState('highlight')
+        const contextAreaFeatureMerged = contextAreaFeatures.reduce(
+          (acc, { geometry, properties }) => {
+            const featureGeometry: Feature<Polygon> = {
+              type: 'Feature',
+              geometry: geometry as Polygon,
+              properties,
+            }
+            if (!acc?.type) return featureGeometry
+            return union(acc, featureGeometry, { properties }) as Feature<Polygon>
```
rdgfuentes, line 108: Is possible that there is an error in this line?

> Argument of type '{ properties: GeoJsonProperties; }' is not assignable to parameter of type 'Feature<Polygon, Properties>'.
>   Type '{ properties: GeoJsonProperties; }' is missing the following properties from type 'Feature<Polygon, Properties>': type, geometryts(2345)
>

rdgfuentes: Besides that, I wonder how would the `union` method merge the properties from multiple polygons. I think it will just leave the `properties` from the last item in the `contextAreaFeatures` array, then it'll use that for the name of the AnalysisGeometry. Is that ok?

applications/fishing-map/src/features/analysis/Analysis.tsx
```diff
@@ -0,0 +1,243 @@
...
+            geometry: contextAreaFeatureMerged,
+            name: name || value,
+          })
+        )
+        const bounds = bbox(contextAreaFeatureMerged) as [number, number, number, number]
+        fitMapBounds(bounds, 10)
+        const featureState = { source: sourceId, sourceLayer: 'main', id: areaId }
+        updateFeatureState([featureState], 'highlight')
+      } else {
+        console.warn('No feature for analysis found')
```
rdgfuentes, line 124: How should the user perceive this scenario? is it possible? what should we display to them?

applications/fishing-map/src/features/analysis/Analysis.tsx
j8seangel: > Is possible that there is an error in this line?

I haven't had that error, did you run `yarn` in the project to ensure all dependencies were updated?

> I think it will just leave the properties from the last item in the contextAreaFeatures array

you are right, I'm just doing it to ensure we don't miss any part of the geometry because the documentation says:

"The domain of the query includes all currently-loaded vector tiles and GeoJSON source tiles: this function does not check tiles outside the currently visible viewport.
Because features come from tiled vector data or GeoJSON data that is converted to tiles internally, feature geometries may be split or duplicated across tile boundaries and, as a result, features may appear multiple times in query results."

but properties should be the same in all the features so we are safe to use it for the name

applications/fishing-map/src/features/analysis/Analysis.worker.ts
```diff
@@ -0,0 +1,39 @@
...
+        simplifiedPoly.type === 'MultiPolygon'
+          ? simplifiedPoly.coordinates.some((coordinates) =>
+              booleanContains({ type: 'Polygon', coordinates }, feature.geometry as Polygon)
+            )
+          : booleanContains(simplifiedPoly, feature.geometry as Polygon)
+
+      if (isContained) {
+        acc.contained.push(feature)
+      } else {
+        // TODO try to get the % of overlapping to use it in the value calculation
```
rdgfuentes, line 26: I doubt getting the % of overlapping and use that in the value calculation would be a better approximation, it could be totally possible that the overlapping area does not match the area of the box where apparent fishing activity is happening since fishing activity in the cell could be not evenly distributed (I think this is the most common)

j8seangel: Indeed, this needs further investigation... we'll need to brainstorm on this

applications/fishing-map/src/features/analysis/Analysis.tsx
j8seangel: It shouldn't ever happen... in the current state because you can't remove any layer from the analysis but thanks for the catch! I've handled the error in case this happens using this:
image

applications/fishing-map/src/features/analysis/Analysis.worker.ts
rdgfuentes: I agree with @enriquetuya, we might evaluate getting the chart data from backend. I think it should provide better accuracy and performance.

applications/fishing-map/src/features/analysis/AnalysisGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+  graphUnit?: string
+}
+
+const tickFormatter = (tick: number) => {
+  const formatter = tick < 1 ? '~r' : '~s'
+  return format(formatter)(tick)
+}
+
+const formatDates = (tick: string) => {
+  const tickDate = DateTime.fromISO(tick)
```
rdgfuentes, line 36: This will display the date in the users's timezone, shouldn't we use always UTC? if so then you should use `DateTime.fromISO(tick).toUTC()`. In my case I see `29 Dec 19` instead of `30 Dec 19` that is the start date of the time range I selected

satellitestudiodesign: ✔️

applications/fishing-map/src/features/analysis/AnalysisGraph.tsx
```diff
@@ -0,0 +1,192 @@
...
+  timeChunkInterval?: Interval
+}
+
+const tickFormatter = (tick: number) => {
+  const formatter = tick < 1 ? '~r' : '~s'
+  return format(formatter)(tick)
+}
+
+const formatDateTicks = (tick: string, timeChunkInterval: Interval) => {
+  const date = DateTime.fromISO(tick).toUTC()
```
rdgfuentes, line 38: :+1:

applications/fishing-map/src/features/analysis/AnalysisGraphWrapper.tsx
```diff
@@ -0,0 +1,151 @@
...
+      const valuesContained = getTimeSeries(
+        (filteredFeatures.contained || []) as any,
+        numSublayers,
+        chunkQuantizeOffset
+      ).map((frameValues) => {
+        return {
+          value: frameValues[0],
+          date: new Date(
+            quantizeOffsetToDate(frameValues.frame, timeChunks.interval).getTime()
+          ).toISOString(),
```
rdgfuentes, line 55: Is this date in UTC? or local timezone of the user? How should be displayed?

applications/fishing-map/src/features/analysis/AnalysisGraphWrapper.tsx
```diff
@@ -0,0 +1,151 @@
...
+      const valuesOverlapping = getTimeSeries(
+        allFilteredFeatues as any,
+        numSublayers,
+        chunkQuantizeOffset
+      ).map((frameValues) => {
+        return {
+          value: frameValues[0],
+          date: new Date(
+            quantizeOffsetToDate(frameValues.frame, timeChunks.interval).getTime()
+          ).toISOString(),
```
rdgfuentes, line 72: Is this date in UTC? or local timezone of the user? How should be displayed?

applications/fishing-map/src/features/analysis/AnalysisGraphWrapper.tsx
```diff
@@ -0,0 +1,151 @@
...
+      const currentDate = DateTime.fromISO(current.date)
+      const startDate = DateTime.fromISO(start)
+      const endDate = DateTime.fromISO(end)
+      return currentDate >= startDate && currentDate < endDate
+    })
+  }, [timeseries, start, end])
+
+  if (!sourceLoaded || generatingTimeseries) return <Spinner className={styles.spinner} />
+
+  if (!timeSeriesFiltered) return <p className={styles.emptyDataPlaceholder}>No data available</p>
```
rdgfuentes, line 140: Translate the empty data message

applications/fishing-map/src/features/analysis/AnalysisLayerPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+type LayerPanelProps = {
+  index: number
+  dataview: UrlDataviewInstance
+}
+
+function AnalysisLayerPanel({ dataview }: LayerPanelProps): React.ReactElement {
+  const { t } = useTranslation()
+  const analysisAreaName = useSelector(selectAnalysisAreaName)
+  const fishignDataview = isFishingDataview(dataview)
```
rdgfuentes, line 26: Minor, typo here in `fishignDataview`

applications/fishing-map/src/features/analysis/AnalysisLayerPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+    {
+      id: 'area',
+      label:
+        analysisAreaName?.length > 35 ? `${analysisAreaName?.slice(0, 35)}...` : analysisAreaName,
+      tooltip: analysisAreaName,
+    },
+  ]
+
+  return (
+    <div className={cx(styles.row)}>
```
rdgfuentes, line 48: Awesome , it's really clear this component markup 👏

applications/fishing-map/src/features/map/controls/MapControls.tsx
```diff
@@ -156,32 +158,36 @@ const MapControls = ({
...
-                }
-                placement="left"
-              >
-                <button
-                  className={cx(styles.basemapSwitcher, styles[currentBasemap])}
-                  onClick={switchBasemap}
-                ></button>
-              </Tooltip>
+              {!isAnalyzing && <Rulers />}
+              {!isAnalyzing && (
```
rdgfuentes, line 162: 👍 Good catch!

applications/fishing-map/src/features/map/map-viewport.hooks.ts
```diff
@@ -1,94 +1,113 @@
...
-  const urlViewport = useSelector(selectViewport)
-  const callback = useCallback((viewport) => dispatchQueryParams(viewport), [dispatchQueryParams])
-  return useDebouncedViewport(urlViewport, callback)
+export function useMapFitBounds() {
+  const mapInstance = useMapboxInstance()
+  const { setMapCoordinates } = useViewport()
+
+  const fitMapBounds = useCallback(
+    (bounds: [number, number, number, number], padding = 60) => {
+      const width = mapInstance ? parseInt(mapInstance.getCanvas().style.width) : 1280
```
rdgfuentes, line 97: I'd use constants for the default values to avoid having magic numbers.

applications/fishing-map/src/features/map/map-viewport.hooks.ts
```diff
@@ -1,94 +1,113 @@
...
-  urlViewport: MapCoordinates,
-  callback: (viewport: MapCoordinates) => void
-): UseViewport {
-  const [viewport, setViewport] = useState<MapCoordinates>(urlViewport)
-  const debouncedViewport = useDebounce<MapCoordinates>(viewport, 400)
 
-  const setMapCoordinates = useCallback((viewport: SetMapCoordinatesArgs) => {
-    setViewport({ ...viewport })
-  }, [])
+const URL_VIEWPORT_DEBOUNCED_TIME = 1000
```
rdgfuentes, line 21: 👍

applications/fishing-map/src/features/map/popups/ContextLayers.tsx
```diff
@@ -114,42 +127,30 @@ type ContextTooltipRowProps = {
...
+
       batch(() => {
-        dispatch(
-          setReportGeometry({
-            geometry,
-            areaName: boundaries.map((b) => b.properties?.geoname ?? []).join(', '),
-          })
-        )
-        dispatchQueryParams({ report: 'fishing-activity' })
+        dispatchQueryParams({
```
rdgfuentes, line 142: 👍

applications/fishing-map/src/features/sidebar/Sidebar.tsx
```diff
@@ -91,8 +92,8 @@ function Sidebar({ onMenuClick }: SidebarProps) {
     return <Search />
   }
 
-  if (reportQuery !== undefined) {
-    return <Report type={reportQuery} />
+  if (isAnalyzing) {
```
rdgfuentes, line 95: 👍

applications/fishing-map/src/features/timebar/Timebar.tsx
```diff
@@ -27,6 +27,26 @@ import {
 } from './timebar.selectors'
 import styles from './Timebar.module.css'
 
+const filterByViewport = (
```
rdgfuentes, line 30: Thanks for extracting this into a function

applications/fishing-map/src/features/workspace/heatmaps/HeatmapLayerPanel.tsx
```diff
@@ -173,65 +151,25 @@ function LayerPanel({ dataview, index, isOpen }: LayerPanelProps): React.ReactEl
...
-                <label>{t('vessel.vesselType_plural', 'Vessel types')}</label>
-                <TagList
-                  tags={vesselsSelected}
-                  color={dataview.config?.color}
-                  className={styles.tagList}
-                />
-              </div>
-            )}
+            <div className={styles.filters}>
+              <DatasetFilterSource dataview={dataview} />
```
rdgfuentes, line 155: 💯

applications/fishing-map/src/features/workspace/shared/DatasetFlagsField.tsx
```diff
@@ -0,0 +1,32 @@
...
+import TagList from '@globalfishingwatch/ui-components/dist/tag-list'
+import { getFlagsByIds } from 'utils/flags'
+import { UrlDataviewInstance } from 'types'
+import styles from 'features/workspace/shared/LayerPanel.module.css'
+
+type DatasetFlagFieldProps = {
+  dataview: UrlDataviewInstance
+}
+
+function DatasetFlagField({ dataview }: DatasetFlagFieldProps) {
```
rdgfuentes, line 12: 👍

packages/fourwings-aggregate/src/get-time-series.ts
```diff
@@ -8,6 +8,10 @@ type Feature = {
 }
 
 const getTimeSeries = (features: Feature[], numSublayers: number, quantizeOffset = 0) => {
+  if (!features || !features.length) {
```
rdgfuentes, line 11: 💯

rdgfuentes (approved):
Great Work 🥇 !

nerik (commented):
This is 💯

applications/fishing-map/src/features/analysis/AnalysisGraphWrapper.tsx
```diff
@@ -0,0 +1,153 @@
...
+            quantizeOffsetToDate(frameValues.frame, timeChunks.interval).getTime()
+          ).toISOString(),
+        }
+      })
+
+      const timeseries = valuesOverlapping.map(({ value, date }) => {
+        const min = valuesContained.find((overlap) => overlap.date === date)?.value
+        return {
+          date,
+          min: min ? min / VALUE_MULTIPLIER : 0,
```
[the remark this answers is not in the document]
j8seangel: Thanks! I didn't know the existence of that method, and I agree on use it instead of exporting the VALUE_MULTIPLIER

applications/fishing-map/src/features/analysis/AnalysisGraphWrapper.tsx
```diff
@@ -0,0 +1,153 @@
+import React, { useEffect, useMemo, useState } from 'react'
+import { useSelector } from 'react-redux'
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import createAnalysisWorker from 'workerize-loader!./Analysis.worker'
+import { Feature, Geometry } from 'geojson'
+import { DateTime } from 'luxon'
+import { getTimeSeries, VALUE_MULTIPLIER } from '@globalfishingwatch/fourwings-aggregate'
```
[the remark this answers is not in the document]
j8seangel: Same than above, thanks for the clarification

j8seangel:
> I'm also wondering, why the need to access the map instance, and then the canvas width, rather than simply querying the containing DOM elements w/h ? 

Used a fallback to get the screen size instead of a fixed value

> what would happen if calling directly Mapbox GL's native fitBounds ?

The latitude and longitude are not update which means when you try to move the map it restores its previous location

applications/fishing-map/src/features/analysis/Analysis.worker.ts
```diff
@@ -0,0 +1,46 @@
...
+type Feature = GeoJSON.Feature<GeoJSON.Geometry>
+export type FilteredPolygons = {
+  contained: Feature[]
+  overlapping: Feature[]
+}
+
+export function filterByPolygon(cells: Feature[], polygon: Geometry): FilteredPolygons {
+  // const t0 = performance.now()
+
+  const simplifiedPoly = simplify(polygon as Polygon | MultiPolygon, { tolerance: 0.1 })
```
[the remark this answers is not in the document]
j8seangel: Good catch! I've moved it and memoized at the previous step so we don't need to recompute on every change.

applications/fishing-map/src/features/analysis/Analysis.worker.ts
```diff
@@ -0,0 +1,46 @@
...
+  const filtered = cells.reduce(
+    (acc, cell) => {
+      const isContained =
+        simplifiedPoly.type === 'MultiPolygon'
+          ? simplifiedPoly.coordinates.some((coordinates) =>
+              booleanContains({ type: 'Polygon', coordinates }, cell.geometry as Polygon)
```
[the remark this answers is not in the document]
j8seangel: Very good idea! I've implemented it here: https://github.com/GlobalFishingWatch/frontend/commit/e1ef45440e98b3b3d126aee0e2cdb6e56063c30b
And the performance went down from `788ms` to `240ms` for the hardest geometry (Chile) and the entire time range:
image

applications/fishing-map/src/features/analysis/AnalysisGraphWrapper.tsx
```diff
@@ -0,0 +1,153 @@
...
+            quantizeOffsetToDate(frameValues.frame, timeChunks.interval).getTime()
+          ).toISOString(),
+        }
+      })
+
+      const allFilteredFeatues = [
+        ...(filteredFeatures.contained || []),
+        ...(filteredFeatures.overlapping || []),
+      ]
+      const valuesOverlapping = getTimeSeries(
```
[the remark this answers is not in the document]
j8seangel: renamed to use something easier to understand like `valuesContainedAndOverlapping` this is something we might can improve in performance as well but I think it is not very important for now

j8seangel:
Thanks everyone for your comments and feedback! It was en entire and complex feature to test 🙇

PR #350, opened 2021-03-05
Updated dataset to use v20201001

rdgfuentes:


applications/vessel-history/src/utils/vessel.test.tsx
```diff
@@ -15,7 +15,7 @@ describe('getVesselAPISource', () => {
     otherShipnames: [],
     shipname: 'DON TITO',
     source: 'AIS',
-    dataset: 'public-global-vessels:v20190502',
+    dataset: 'public-global-fishing-vessels:v20201001',
```
enriquetuya, line 18: Is this a unit test that mocks the API or you are actually calling the service?
If it is a mock you could change this to `vYYYYMMDD` so it is clear it is a mock.

applications/vessel-history/src/features/vessel-list-item/VesselListItem.test.tsx
```diff
@@ -23,7 +23,7 @@ describe('<VesselListItem />', () => {
     otherShipnames: [],
     shipname: 'DON TITO',
     source: 'AIS',
-    dataset: 'public-global-vessels:v20190502',
+    dataset: 'public-global-fishing-vessels:v20201001',
```
enriquetuya, line 26: Is this a unit test that mocks the API or you are actually calling the service?
If it is a mock you could change this to `vYYYYMMDD` so it is clear it is a mock.

enriquetuya (commented):
@rdgfuentes Looks good to me. I've added a couple of comments in the tests but nothing critical.

PR #354, opened 2021-03-09
fixed low level of activity at the bottom of the tiles

nerik:
Reported here: https://globalfishingwatch.atlassian.net/browse/MAP-285

Before:
Screenshot 2021-03-09 at 17 10 51

After: 
Screenshot 2021-03-09 at 17 04 04


j8seangel:
very good catch! it LGTM and thanks for adding the test!

PR #357, opened 2021-03-11
Fishingmap/default map context component

j8seangel:
Remove custom implementation of the mapInstance hook to use the recently discovered [with in the react-map-gl library](https://github.com/visgl/react-map-gl/blob/master/docs/api-reference/map-context.md)

As bonuses I replace the fitBounds custom implementations to reuse the `useMapFitBounds`


applications/fishing-map/src/features/app/App.tsx
```diff
@@ -125,7 +125,8 @@ function App(): React.ReactElement {
 
   return (
     <Fragment>
-      <MapboxRefProvider>
+      {/* Value as null as there is no needed to set a default value but Typescript complains */}
+      <MapContext.Provider value={null as any}>
```
nerik, line 129: Is there any reason to not have the `Provider` at the top level, just below the Fragment?

applications/fishing-map/src/features/map/Map.tsx
```diff
@@ -223,30 +211,28 @@ const MapWrapper = (): React.ReactElement | null => {
...
+      map.showTileBoundaries = debugOptions.debug
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [mapRef.current])
+  }, [map, debugOptions])
 
   return (
     <div className={styles.container}>
-      {<MapScreenshot map={mapRef.current?.getMap()} />}
+      {<MapScreenshot map={map} />}
```
nerik, line 232: 👍

applications/fishing-map/src/features/map/Map.tsx
```diff
@@ -138,8 +131,7 @@ const getLegendLayers = (
 }
 
 const MapWrapper = (): React.ReactElement | null => {
-  const mapRef = useMapboxRef()
-  const mapInstance = useMapboxInstance()
+  const map = useMapInstance()
```
nerik, line 134: Minor, maybe I'll fix it later but naming this `map` is confusing to me. 

What is `map`? 
- the Map component in Map.tsx?
- the Map component from RMGL?
- the Mapbox GL map instance?

I'd rather have
```
const mapboxGLInstance = useMapboxGLInstance()
```

applications/fishing-map/src/features/map/controls/MapSearch.tsx
```diff
@@ -2,36 +2,24 @@ import React, { memo, useRef, useState } from 'react'
...
   const inputRef = useRef<HTMLInputElement | null>(null)
   const [areasMatching, setAreasMatching] = useState<OceanArea[]>([])
-  const mapInstance = useMapboxInstance()
 
-  const { setMapCoordinates } = useViewport()
+  const fitBounds = useMapFitBounds()
 
   const onSelectResult = ({ selectedItem }: UseComboboxStateChange<OceanArea>) => {
-    const bounds = selectedItem?.properties.bounds
+    const bounds = selectedItem?.properties.bounds as [number, number, number, number]
```
nerik, line 20: Minor, I checked and there are already 8 occurrences of `[number, number, number, number]` in `fishing-map`. I suggest we use 
```
import { BBox } from 'geojson'
``` 
What do you think?

nerik (commented):
This is much cleaner, thank you!
I only have minor comments that can be addressed later.
🙌

applications/fishing-map/src/features/map/controls/MapSearch.tsx
j8seangel: I had to do it because the type definition of bbox in geojson package is
```
export type BBox = [number, number, number, number] | [number, number, number, number, number, number];
```

that is way I needed to define which of both formats we are using, maybe we can just create our own definition of BBox as store in `/types`  WDYT ?

applications/fishing-map/src/features/app/App.tsx
j8seangel: Yep, the Menu and the Modal components are "siblings" so we needed a Fragment parent component

applications/fishing-map/src/features/map/Map.tsx
j8seangel: I like it, I'll keep it in mind for next PRs

applications/fishing-map/src/features/app/App.tsx
nerik: I mean get rid of Fragment and
```
<Provider>
  <SplitView />
  <Menu />
  <Modal />
  ....
</Provider>

applications/fishing-map/src/features/map/controls/MapSearch.tsx
nerik: Perfect

applications/fishing-map/src/features/app/App.tsx
j8seangel: Oh I get it! sounds good, I've updated it in https://github.com/GlobalFishingWatch/frontend/commit/e809883567fb188f663c85a825c94bfe11e9df73

applications/fishing-map/src/features/map/controls/MapSearch.tsx
j8seangel: 👍 https://github.com/GlobalFishingWatch/frontend/commit/7ebe184dfc4513b7cb1f50503202ad3c23103997

PR #368, opened 2021-03-16
[WIP] Fishingmap/vessel events

nerik:
- [x] Fetch events on interaction
- [x] Show events in timebar
- [x] Show events in map 
- [x] Filter event types by the timebar settings selector
- [x] Interaction
  - [x] Tooltips in timebar and map
  - [x] Click in timebar fits geographically 
  - [x] ~Click in map fits temporarily~
- [x] Events legend 
- [x] Remove ad-hoc code for events parsing in resource.slice once the API does it
- [x] Review performance and date filtering
- [ ] Support multiple events

⚠️ Some packages are deployed to npm directly from this branch to make the preview available for data validation 🙏 take care before merging on versions conflicts

image


rdgfuentes:
@nerik I know this is WIP but I looked a bit into it and I thought it is a good case where we can implement some tests for the `HeatmapTooltipRow` component using the `datasets.mock`, testing also the dataview instance is upserted when the user clicks the vessel button, not only for adding coverage to the  component but also to help us on during the development (kind of TDD). I created this PR #372  as a start point for it, I hope it helps!

packages/layer-composer/src/generators/types.ts
```diff
@@ -232,6 +234,9 @@ export interface TrackGeneratorConfig extends GeneratorConfig {
 export interface VesselEventsGeneratorConfig extends GeneratorConfig {
   type: Type.VesselEvents
   data: RawEvent[]
+  color?: string
+  track?: any
+  showTrackSegments?: boolean
```
nerik, line 239: As far as I can see this is never practically set to false. I would remove configurability as long as it's not needed.

packages/layer-composer/src/generators/vessel-events/vessel-events.ts
```diff
@@ -68,39 +51,66 @@ class VesselsEventsGenerator {
...
-    const source: VesselsEventsSource = {
-      id,
+    const pointsSource: VesselsEventsSource = {
+      id: `${id}_points`,
       type: 'geojson',
       data: newData,
     }
-    return [source]
+
+    if (!config.showTrackSegments) {
```
nerik, line 60:
```suggestion
    if (!config.zoom <= POINTS_TO_SEGMENTS_ZOOM_LEVEL_SWITCH) {
```
This is one of the potential issues with performance. Segments are calculated even though zoom allows only for displaying points.

packages/layer-composer/src/generators/vessel-events/vessel-events.ts
```diff
@@ -110,15 +120,44 @@ class VesselsEventsGenerator {
         metadata: {
           group: Group.Point,
         },
-      },
+      } as SymbolLayer,
+    ]
+    if (!config.showTrackSegments) {
```
nerik, line 125:
```suggestion
    if (!config.zoom <= POINTS_TO_SEGMENTS_ZOOM_LEVEL_SWITCH) {
```

packages/layer-composer/src/generators/vessel-events/vessel-events.ts
```diff
@@ -68,39 +51,66 @@ class VesselsEventsGenerator {
...
+      return [pointsSource]
+    }
+    // TODO review performance memoization
+
+    const segments = memoizeCache[config.id].getVesselSegmentsGeojson(
+      track,
+      data,
+      start,
+      end,
+      currentEventId
```
nerik, line 70: Another potential performance fix on timebar hover: segments GeoJSON could be first generated entirely. Then another pass to attach the `active` property to geojson features. This way `getVesselSegmentsGeojson` can be memoized without depending on `currentEventId`.

nerik: It might even be better to do:
- do not attach `active` property
- modify expression here : https://github.com/GlobalFishingWatch/frontend/pull/368/files#diff-3a95f5df66ac12142259807dff47e753fd374a82483431c24ebb6dfbcef048fdR86
  to directly compare `event.id` and `currentEventId`

packages/layer-composer/src/generators/vessel-events/vessel-events.utils.ts
```diff
@@ -0,0 +1,139 @@
...
+        type: 'Point',
+        coordinates: [lng, event.position.lat],
+      },
+    }
+  })
+
+  return featureCollection
+}
+
+export const getVesselSegmentsGeojson = (
```
nerik, line 97: Another performance possible improvement:

Instead of doing, every time timebar start/end changes:
- filter entire track by timebar start/end
- loop through all events and filter entire track with event start/end

It would be more efficient to split the process in two stages:
- loop through all events and filter entire track with event start/end, just ONCE
- then filter resulting event segments by timebar start/end when timebar start/end changes

What do you think?

nerik (commented):
This is looking spectacular 👏 

I think we should merge this as it is, and discuss together possible perf improvements for implementation in another PR. Although it should be adressed before release IMHO.

packages/layer-composer/src/generators/types.ts
j8seangel: Indeed, this is going to be removed

packages/layer-composer/src/generators/vessel-events/vessel-events.ts
j8seangel: Good catch! I'll replace it with 
```js
(!config.track || (config.zoom as number) <= SEGMENTS_MIN_ZOOM_LEVEL)
```
this way we ensure we calculate the segments only when necessary and when generator has data

packages/layer-composer/src/generators/vessel-events/vessel-events.utils.ts
j8seangel: Good idea! Implemented in https://github.com/GlobalFishingWatch/frontend/pull/368/commits/5fdf4c30bdc8e6bc446f85c7160b0aa45d861da5

packages/layer-composer/src/generators/vessel-events/vessel-events.ts
j8seangel: 👏 it works way better! see: https://github.com/GlobalFishingWatch/frontend/pull/368/commits/e22723264fdc42d16d3603fd6421d2caf774a1d2

j8seangel:
Thanks for the review! the performance has increased a lot!

j8seangel:
Merging and I'll address the `support for multiple events` in another PR

PR #371, opened 2021-03-18
Use client-side aggregation with env layers

nerik:
https://user-images.githubusercontent.com/1583415/112144290-49f6b080-8bd9-11eb-8b37-807fbb014f04.mov

- Added a `TileAggregationParams` type and renamed some of its parameters to clarify
- Allow multiple heatmap animated layers
- Allow distinct heatmap animated layers to work with different timechunk intervals (defaults to `auto` which is the current fishing layers behavior)
- Added a `month` timechunk interval
- Adds a `none` `SublayerCombinationMode` (how values are aggregated _between sublayers_), which actually does nothing, to be used when there's a single sublayer  
- Added an `aggregationMode` (how values are aggregated _over time_) param which in addition to the already existing `add`, adds an `avg` mode.   (+tests)
- Added configurability for the value multiplier since currently environment layer values come without it

This PR does not includes `fishing-map` functionality yet.
This PR depends on MGL updated done here See https://github.com/GlobalFishingWatch/mapbox-gl-js/pull/110


packages/fourwings-aggregate/src/aggregate-cell.ts
```diff
@@ -6,7 +8,8 @@ const aggregateCell = (
   delta: number,
   quantizeOffset: number,
   numSublayers: number,
-  debug = false
+  aggregationOperation = AggregationOperation.sum,
+  multiplier = VALUE_MULTIPLIER
```
j8seangel, line 12: This https://github.com/GlobalFishingWatch/frontend/blob/fourwings-aggregate/env-layers/packages/react-hooks/src/use-map-interaction/use-map-interaction.ts#L44 is still using the old params format order.
At least we need to remove the `debug` param to use the default ones.

Which points me that there are a lot of params, and it enforces to use the not required ones at the end. What about moving to a configuration object and destructure like:
```js
const aggregateCell = ({
  rawValues: string,
  frame: number,
  delta: number,
  quantizeOffset: number,
  numSublayers: number,
  aggregationOperation: AggregationOperation.sum,
  multiplier: VALUE_MULTIPLIER
}) => {
```
I like it because when using it the order doesn't matter anymore and it is easier to read what the function params means:
```js
const values = aggregateCell(
  rawValues: properties.rawValues,
  frame,
  delta: timeChunks.deltaInIntervalUnits,
  quantizeOffset: activeTimeChunk.quantizeOffset,
  numSublayers,
)
```

packages/fourwings-aggregate/src/aggregate-cell.ts
```diff
@@ -28,16 +31,22 @@ const aggregateCell = (
...
   // One aggregated value per sublayer
-  const aggregatedValues = new Array(numSublayers).fill(0)
+  let aggregatedValues = new Array(numSublayers).fill(0)
 
+  let numValues = 0
   for (let i = 0; i < rawValuesArrSlice.length; i++) {
     const sublayerIndex = i % numSublayers
     const rawValue = rawValuesArrSlice[i]
-    if (rawValue) {
+    if (rawValue !== null && rawValue !== undefined) {
```
j8seangel, line 40: good catch, probably you had a lot of 0️⃣ not being used here

packages/fourwings-aggregate/src/aggregate-tile.ts
```diff
@@ -33,13 +40,8 @@ const getCellCoords = (tileBBox: any, cell: number, numCols: number) => {
...
 
-const getPointFeature = (
-  tileBBox: any,
-  cell: number,
-  numCols: number,
-  numRows: number,
-  addMeta: boolean
-): any => {
+const getPointFeature = (featureParams: FeatureParams): any => {
+  const { tileBBox, cell, numCols, numRows, addMeta } = featureParams
```
j8seangel, line 44: This is the same to the previous comment, it looks easier for me, WDYT ?

packages/fourwings-aggregate/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
+export type BBox = [number, number, number, number]
```
j8seangel, line 1: 👏 🥇 thanks!!

packages/fourwings-aggregate/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export enum GeomType {
+  point = 'point',
+  rectangle = 'rectangle',
+}
+
+export enum SublayerCombinationMode {
```
j8seangel, line 8: 👏 🥇 thanks!!

packages/fourwings-aggregate/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  // When set to true, generate a second sourceLayer that contains the raw data
+  interactive: boolean
+  // Sets value breaks for each sublayer. When set, values will return a bucket index instead of the actual value
+  sublayerBreaks?: number[][]
+  // Sets the number of sublayers
+  sublayerCount: number
+  // Defines how sublayers computed values are combined in cell properties
+  sublayerCombinationMode: SublayerCombinationMode
+  // Defines sublayers visibilty, ie aallow toggling sublayers without reloading source
+  sublayerVisibility: boolean[]
```
j8seangel, line 46: If we move to identify sublayers by ids, I think would be useful to use them here as well.

packages/layer-composer/src/generators/heatmap/config.ts
```diff
@@ -9,17 +9,16 @@ export const API_ENDPOINTS = {
...
-  [HeatmapAnimatedMode.Compare]: HeatmapAnimatedCombinationMode.Compare,
-  [HeatmapAnimatedMode.Bivariate]: HeatmapAnimatedCombinationMode.Bivariate,
-  [HeatmapAnimatedMode.Blob]: HeatmapAnimatedCombinationMode.Literal,
-  [HeatmapAnimatedMode.Extruded]: HeatmapAnimatedCombinationMode.Cumulative,
+export const HEATMAP_MODE_COMBINATION: Record<HeatmapAnimatedMode, SublayerCombinationMode> = {
+  [HeatmapAnimatedMode.Single]: SublayerCombinationMode.none,
+  [HeatmapAnimatedMode.Compare]: SublayerCombinationMode.max,
+  [HeatmapAnimatedMode.Bivariate]: SublayerCombinationMode.bivariate,
+  [HeatmapAnimatedMode.Blob]: SublayerCombinationMode.literal,
+  [HeatmapAnimatedMode.Extruded]: SublayerCombinationMode.cumulative,
```
j8seangel, line 17: Not sure if I missed something yesterday from conversation about enums.
Are we using Enum.Property in uppercase or lowercase ? just to keep the same pattern for next things

packages/layer-composer/src/generators/heatmap/util/index.ts
```diff
@@ -6,3 +9,15 @@ export const toURLArray = (paramName: string, arr: string[]) => {
     })
     .join('&')
 }
+
+export const getSourceId = (config: GlobalHeatmapAnimatedGeneratorConfig, timeChunk: TimeChunk) => {
+  return `${config.id}-${timeChunk.id}`
+}
+
+export const getLayerId = (
```
j8seangel, line 17: I love it, thanks for create it! 
However I'd like to talk about normalise the separators, because for example in [context-layers](https://github.com/GlobalFishingWatch/frontend/blob/fourwings-aggregate/env-layers/packages/layer-composer/src/generators/context/context-layers.ts#L104) it is using `_` instead of '-' which ends up having a hard to read sources /layers list. So let's choose one and use it everywhere 🙏

packages/fourwings-aggregate/src/types/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  // Number of rames to aggregate, ie number of frames between active time start & end
+  delta: number
+  // Which type of geometry to generate
+  geomType: GeomType
+  // When set to true, aggregation will be done on the backend
+  singleFrame: boolean
+  // When set to true, generate a second sourceLayer that contains the raw data
+  interactive: boolean
+  // Sets value breaks for each sublayer. When set, values will return a bucket index instead of the actual value
+  sublayerBreaks?: number[][]
```
j8seangel, line 40: should we force to use a fixed length array ?

packages/layer-composer/src/generators/heatmap/util/time-chunks.ts
```diff
@@ -190,18 +206,20 @@ export const getActiveTimeChunks = (
   activeStart: string,
   activeEnd: string,
   datasetStart: string,
-  datasetEnd: string
+  datasetEnd: string,
+  interval_?: IntervalOption
```
j8seangel, line 210: any reason to have the _ suffix ?

j8seangel (commented):
The typings are great! This is super useful to consume and also to use as documentation!
There are several minor comments but in general it LGTM, great work!

Let's talk when we move to use ids instead of indexes for sublayers because it will have a lot of implications on fishing-map but looking forward to see this working there.

packages/fourwings-aggregate/src/types/index.ts
nerik: Let me check. Thing is, since we pass these parameters via query params, there's not so much flexibility in nesting. Although I'm think we could simply use `qs` the same way we do for the URL workspace.

packages/layer-composer/src/generators/heatmap/util/index.ts
nerik: Go for `_`

packages/layer-composer/src/generators/heatmap/util/time-chunks.ts
nerik: None, will fix

packages/fourwings-aggregate/src/types/index.ts
nerik: Ah I misunderstood your comment. I'm not sure we need to move to identify sublayers by id. Using sublayer indices within HeatmapAnimated layers wwith ids should still work fine

packages/layer-composer/src/generators/heatmap/config.ts
j8seangel: Whatever you prefer, both looks to me.
I'd say uppercase as there are lot of places already using it

PR #375, opened 2021-03-23
Fishingmap/tracks thinning

satellitestudiodesign:
Send thinning parameters to the tracks API to avoid downloading unnecessary points.

There are 2 levels
- Guest users will see an aggressively thinned track.
- Logged in users will see a thinned track that is visually almost unnoticeably different from the original

Left: Logged user (678kb) / Right: original resolution (6.8M)
Screenshot 2021-03-23 at 19 38 33
Same example in aggressive mode returns around 280kb track size

There's also an additional debug param to disable thinning completely.
Screenshot 2021-03-23 at 19 42 35


nerik:
Very cool 👏 
One comment is that ideally, settings for `Default` and `Aggressive` thinning should live in distinct track dataviews, with maybe a third profile for no thinning at all (the one currently when toggling off thinning in debug mode).
Thoughts? @j8seangel @satellitestudiodesign ?

applications/fishing-map/src/data/config.ts
```diff
@@ -34,6 +34,34 @@ export const DEFAULT_WORKSPACE = {
   version: DEFAULT_VERSION,
 }
 
+export enum ThinningLevels {
+  Aggresive = 'aggresive',
```
nerik, line 38:
```suggestion
  Aggressive = 'aggressive',
```

applications/fishing-map/src/features/resources/resources.slice.ts
```diff
@@ -26,14 +30,24 @@ const initialState: ResourcesState = {}
...
+    const { thinning } = selectDebugOptions(state)
     const isTrackResource = resource.datasetType === DatasetTypes.Tracks
     const responseType =
       isTrackResource &&
       resource.datasetConfig.query?.some((q) => q.id === 'binary' && q.value === true)
         ? 'vessel'
         : 'json'
-    const data = await GFWAPI.fetch(resource.url, { responseType }).then((data) => {
+    let url = resource.url
+    if (isTrackResource && thinning) {
```
nerik, line 44: See my dataviews comment - the problem here is that we introduce track-specific logic in a generic resource thunk

applications/fishing-map/src/features/resources/resources.selectors.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ResourceQuery } from 'features/resources/resources.slice'
+import { selectDataviewInstancesResolved } from 'features/workspace/workspace.selectors'
+import { UrlDataviewInstance } from 'types'
+import { isGuestUser } from 'features/user/user.selectors'
+import { selectDebugOptions } from 'features/debug/debug.slice'
+import { ThinningLevels, THINNING_LEVELS } from 'data/config'
+
+export const resolveDataviewDatasetResource = (
+  dataview: UrlDataviewInstance,
+  { type, id }: { type?: DatasetTypes; id?: string }
```
nerik, line 19: I know that code is not new, but it's hard to understand just looking at the function signature that it's either `type` OR `id`

```suggestion
 typeOrId: DatasetTypes | string```
 
 Then, I guess we should look at https://medium.com/ovrsea/checking-the-type-of-an-object-in-typescript-the-type-guards-24d98d9119b0

applications/fishing-map/src/features/resources/resources.slice.ts
j8seangel: Agreed on this should be dataview responsibility instead of doing it in the thunk, it has been updated
Also there is a [new enum](https://github.com/GlobalFishingWatch/frontend/pull/375/commits/d99bf5f30bceed021ac990a0de8b4e992189be73#diff-d1dbd13d07dbb0113e847e8262d6284320eee1a5c5fb1ce2d55a2a31f21e000dR17) to better use of the API dataset endpoints

PR #376, opened 2021-03-24
URL workspace preview window

nerik:
Screenshot 2021-03-24 at 16 33 05

Adds a URL workspace preview window in the debug modal.

Additionally, shortens the URL by replacing common terms by abbreviations.


j8seangel (approved):
💯 Perfect!

PR #395, opened 2021-03-29
Packages/dataview client & react-hooks

j8seangel:
Bring back to live the `dataviews-client` package to move the fishing-map logic there to resolve:
- [x] Dataview and dataview instances
- [x] Resolved dataviews to generators
- [x] use-map-legend hook

Also reuses the `Resource` definition from api-types instead of duplicating it in fishing-map


nerik (approved):
This is much cleaner, thanks! Did corrections according to my own comments

PR #421, opened 2021-04-08
VH-33 Offline Map Prototype - Initial basic map & service workers setup

rdgfuentes:
- Initial basic map setup on vessel profile page (only basemap, no events, no tracks yet)
- Service workers setup to cache resources for offline access (translations, images, fonts, css, api requests, auth requests)


applications/vessel-history/src/data/config.ts
```diff
@@ -0,0 +1,137 @@
...
+export const BACKGROUND_LAYER = [
+  {
+    id: 'background',
+    tileset: 'background',
+    description: 'background',
+    type: Generators.Type.Background,
+    color: MAP_BACKGROUND_COLOR,
+  } as BackgroundGeneratorConfig,
+]
+export const DEFAULT_DATAVIEWS = [
```
j8seangel, line 21: This data views format is deprecated, we don't want to use the carts generator anymore.
We could use this list: https://github.com/GlobalFishingWatch/frontend/blob/master/applications/fishing-map/src/data/dataviews.md#context

applications/vessel-history/src/features/map/Map.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { useMapboxRef, useMapboxRefCallback } from './map.context'
+import styles from './Map.module.css'
+
+const Map = (): ReactElement => {
+  const mapRef = useMapboxRef()
+
+  const onRefReady = useMapboxRefCallback()
+  // const { viewport, onViewportChange } = useViewport()
+
+  const generatorConfigs = useSelector(getLayerComposerLayers)
```
j8seangel, line 27: We created a [dataview-client package](https://github.com/GlobalFishingWatch/frontend/blob/master/packages/dataviews-client/src/resolve-dataviews-generators.ts) to convert from dataviews to generators. Let's use it here to simplify the workflow

applications/vessel-history/src/features/map/Map.tsx
```diff
@@ -0,0 +1,85 @@
...
+          latitude={viewport.latitude}
+          longitude={viewport.longitude}
+          zoom={viewport.zoom}
+          // onLoad={onLoadCallback}
+          // onViewportChange={onViewportChange}
+          // mapStyle={styleWithArrows}
+          // onClick={handleMapClick}
+          // onHover={handleMapHover}
+          // onMouseMove={onMapMove}
+          mapOptions={{
```
j8seangel, line 62: Just to not forget, it is better to declare the config as a const outside of the render to avoid re-rendering the entire map on every change as it is a new object instance

applications/vessel-history/src/features/map/map.context.tsx
```diff
@@ -0,0 +1,67 @@
+import React, {
```
j8seangel, line 1: I realised a couple of few days ago the react-map-gl library already offers a context component to read map properties out of the map.
I'd use it here instead of the custom implementation we did, ref: https://github.com/GlobalFishingWatch/frontend/blob/master/applications/fishing-map/src/features/map/map-context.hooks.ts

applications/vessel-history/src/features/map/map.hooks.ts
```diff
@@ -0,0 +1,165 @@
...
+//     if (debouncedViewport && callback) {
+//       callback(debouncedViewport)
+//     }
+//     // eslint-disable-next-line react-hooks/exhaustive-deps
+//   }, [debouncedViewport])
+
+//   return { viewport, onViewportChange, setMapCoordinates }
+// }
+
+// export function useViewport(): UseViewport {
```
j8seangel, line 132: This always was an 🤯 let's talk about the latest working solution I found using [an atom](https://github.com/GlobalFishingWatch/frontend/blob/master/applications/fishing-map/src/features/map/map-viewport.hooks.ts#L26)

j8seangel (commented):
See you at 3pm! lot of things to discuss but this is good to starting

applications/vessel-history/src/data/config.ts
rdgfuentes: Thanks for the advise @j8seangel . We'll definitely do it that way when we implement the final map, this is a super simplified version to have available so that we can prototype the offline access mode

applications/vessel-history/src/features/map/map.hooks.ts
rdgfuentes: I'll try that

applications/vessel-history/src/service-worker.ts
```diff
@@ -0,0 +1,163 @@
...
+      }),
+    ],
+  })
+)
+
+// Cache API Auth Calls
+registerRoute(
+  ({ url }) =>
+    process.env.REACT_APP_API_GATEWAY &&
+    url.origin === process.env.REACT_APP_API_GATEWAY &&
```
j8seangel, line 119: to avoid duplicating this in the auth and the regular API calls what about having a global const like
```const API_GATEWAY = process.env.REACT_APP_API_GATEWAY```
and we check it here and [here](https://github.com/GlobalFishingWatch/frontend/pull/421/files#diff-d2f15e6637529cd54a564a392ebcf7aac0d10c85c4266eefbcb0017d2e1972f7R138) it?

j8seangel (approved):
There seems to be a conflict with the store.ts file the rest looks good to me, amazing work!

fertandil87 (approved):

PR #429, opened 2021-04-12
Ocean areas/i18n

j8seangel:
Support i18n in ocean-areas packages
image


j8seangel:
Merging as I need to setup the crowding integration with these new files

PR #434, opened 2021-04-12
support adding presence layers

j8seangel:
Support adding presence layers in the activity section:
- [x] Show presence on add layer only when dataview included in workspace
- [x] [Auto detect](https://github.com/GlobalFishingWatch/frontend/pull/434/files#diff-f9708754c8ace51fecd0d6922da43e4dfa2044b5729a1f605a99b98043104df7R29) interval reading dataset.configuration.resolution when existing

https://user-images.githubusercontent.com/10500650/114438841-9033a400-9bc8-11eb-8841-d89475e6cce7.mp4


packages/layer-composer/src/generators/heatmap/util/time-chunks.ts
```diff
@@ -212,11 +219,11 @@ export const getActiveTimeChunks = (
   activeEnd: string,
   datasetStart: string,
   datasetEnd: string,
-  interval?: IntervalOption
+  interval?: Interval | Interval[]
 ): TimeChunks => {
   const delta = +toDT(activeEnd) - +toDT(activeStart)
   const finalInterval: Interval =
-    !interval || interval === 'auto' ? getInterval(delta) : (interval as Interval)
+    !interval || Array.isArray(interval) ? getInterval(delta, interval) : (interval as Interval)
```
nerik, line 226: I'm in favor of removing support for `interval` as a single string - we don't need backwards compatibility as resolveDataviews will always return an array anyways (also rename it to `intervals`)

nerik (changes requested):
Apart from one comment and what we reviewed together, all good! Thanks!

packages/layer-composer/src/generators/heatmap/util/time-chunks.ts
j8seangel: We need to support a fixed interval for the environmental layers and moving to an array will make the logic harder so let's go as we are for now and let's keep an eye on this

PR #438, opened 2021-04-13
VH-33 Offline access for vessels & offline map enhancements (prototype)

rdgfuentes:
- Vessels can be saved (and removed) to access them offline
- Added low res layer as basemap to use as fallback when offline if some tiles have not been cached by the service worker yet


applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
```diff
@@ -88,12 +91,12 @@ const VesselListItem: React.FC<ListItemProps> = (props): React.ReactElement => {
             '-'
           )}
         </div>
-        {/* {props.saved && (
+        {props.saved && (
           <div>
             <label>{t('vessel.savedOn', 'saved on')}</label>
-            2020/08/01
+            {`${formatI18nDate(props.saved, { format: DateTime.DATETIME_MED })}`}
```
j8seangel, line 97: 👏

applications/vessel-history/src/index.css
```diff
@@ -1,13 +1,20 @@
...
+
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
+    'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+code {
```
j8seangel, line 18: I like this, do you think worth it to move to the ui-components base styles ?

applications/vessel-history/src/service-worker.ts
```diff
@@ -64,6 +64,9 @@ self.addEventListener('message', (event) => {
 
 // Any other custom service worker logic can go here.
 
+// Precache GeoJson for offline use
+precacheAndRoute([{ url: '/data/ne_10m_admin_0_countries_gj.geojson' }])
```
j8seangel, line 68: If feasible I'd share the path here and within the generator to ensure if the file changes it is updated in both sides

j8seangel (commented):
I missed a spinner while the vessel information loads here:
image

Also it is a bit weird for me to have this label in the button: 
image
but clicking on it behaves like the trash to remove the vessel from offline

In general it looks good! I like the approach of storing the vessels to ensure we have better control over them... 
but if the needs of storing more things offline grows I'm not sure how scalable this is and maybe you can take a look at [redux-persist](https://github.com/rt2zz/redux-persist) to automatically save the entire store

j8seangel:
Also, please remember to clean some warnings
image

applications/vessel-history/public/locales/source/translations.json
```diff
@@ -31,6 +31,8 @@
     "mmsi": "mmsi",
     "noAuthorizations": "No authorizations found",
     "plusPreviousValuesByField": "+{{quantity}} previous {{fieldLabel}}",
+    "readyForOfflineView": "READY FOR OFFLINE VIEW",
+    "remove": "Remove",
```
fertandil87, line 35: remove show be inside common?

applications/vessel-history/src/features/home/Home.tsx
```diff
@@ -27,28 +30,34 @@ interface LoaderProps {
...
   const minimumCharacters = 3
   const resultsPerRequest = 25
 
+  useEffect(() => {
+    dispatchFetchOfflineVessels()
+  }, [dispatchFetchOfflineVessels])
+
   const fetchData = useCallback(async (query: string) => {
-    setSearching(true)
+    setLoading(true)
```
fertandil87, line 48: this change wasn't necessary, will disappear in with my PR

applications/vessel-history/src/features/vessel-list-item/VesselListItem.tsx
fertandil87: I was thinking the same

fertandil87 (changes requested):

applications/vessel-history/src/features/profile/components/Info.module.css
```diff
@@ -45,3 +46,30 @@
...
 }
+.actions {
+  display: flex;
+  justify-content: center;
+}
+.readyButton {
+  flex: 1;
+  text-transform: uppercase;
+  color: var(--color-primary-green) !important;
+  background-color: var(--color-terthiary-green) !important;
```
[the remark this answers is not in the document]
rdgfuentes: Right @j8seangel , the problem is that the button has two classes and the ui component defines a style with both classes selector. As a result the styles in the ui-component has more specificity than the styles that we can define for the component so they are overwritten.

applications/vessel-history/src/service-worker.ts
rdgfuentes: Done

applications/vessel-history/src/index.css
rdgfuentes: Sure we can do it in a separate PR, WDYT @j8seangel ?

applications/vessel-history/src/data/config.ts
```diff
@@ -17,7 +21,32 @@ export const BACKGROUND_LAYER = [
...
+      {
+        type: 'fill',
+        paint: {
+          'fill-color': DEFAULT_LANDMASS_COLOR,
+          'fill-opacity': 0.99, // This is a trick to allow proper rendering of MGL heatmap layers behind
+        },
+      } as FillLayer,
+    ],
+    metadata: {
+      offline: true,
```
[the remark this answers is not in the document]
rdgfuentes: There's no switch yet. Given that this layer is already cached by the sw it's rendered behind the landmass tiles, so that when the landmass tiles (high res) are not available this layer displays the landmass

rdgfuentes:
@j8seangel @fertandil87 I made most of the changes suggested by you. Could you please take a loo

> Also, please remember to clean some warnings
> image

Fixed in upstream branch and merged

applications/vessel-history/src/features/profile/components/Info.module.css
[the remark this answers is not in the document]
j8seangel: When I had the same I created a parent class to force having one more selector weight, something like:
```
.actions .readyButton {
...
```
should be enough to ensure it and avoid the !important selector. not very important in fact

applications/vessel-history/src/data/config.ts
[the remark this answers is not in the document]
j8seangel: I see! it is always there and the cache is done by default! Thanks for the clarification.

applications/vessel-history/src/index.css
j8seangel: Dale!

fertandil87 (approved):

fertandil87:
@rdgfuentes see conflicts

PR #458, opened 2021-04-26
Layer composer/heatmap buckets

j8seangel:
Use buckets api endpoint to generate the color ramp in 4 wings layers:
- [x] Fetch data
- [x] Update layer composer to use fetched data
- [ ] Abort previous request on changes to avoid race conditions  
- [ ] Request only visible datasets
- [ ] Loading spinner while fetching bins
- [ ] Include filters 
- [ ] Bivariate mode
- [ ] Blob & extruded mode

Fixes requesting the tiles twice when loading as we used the default timebar instead of waiting for the workspace


packages/dataviews-client/src/resolve-dataviews-generators.ts
```diff
@@ -46,6 +46,32 @@ type DataviewsGeneratorConfigsParams = {
...
+      date.getFullYear(),
+      date.getMonth(),
+      date.getDate(),
+      date.getHours(),
+      date.getMinutes()
+    )
+  )
+}
+
+const getDatasetsExtent = (datasets: Dataset[] | undefined) => {
```
nerik, line 62: Nice 👍
Would be actually cool to have that in a selector and send to Timebar's `absoluteStart|End` instead of using hardcoded `DEFAULT_WORKSPACE`

packages/layer-composer/src/generators/gl/gl.ts
```diff
@@ -9,16 +8,12 @@ class GlStyleGenerator {
   }
 
   _getStyleLayers = (config: GlGeneratorConfig) => {
-    const layout = {
-      visibility: isConfigVisible(config),
```
nerik, line 13: Thanks!

packages/layer-composer/src/generators/heatmap/util/fetch-breaks.ts
```diff
@@ -0,0 +1,73 @@
...
+    }),
+  })
+    .then((r) => {
+      if (r.ok) return r.json()
+      throw r
+    })
+    .then((breaks: Breaks) => {
+      const max = Math.max(...breaks.flatMap((b) => b))
+      const maxBreaks = breaks.find((b, index) => b[breaks[index].length - 1] === max) || breaks[0]
+      // We want to use the biggest break in every sublayer
```
nerik, line 72: Mental note to modify that when we'll need different scales in bivariate mode.

nerik (approved):
Mwah 💅

PR #459, opened 2021-04-26
MAP-329 Create single analysis report on multiple dataviews

rdgfuentes:
- [x] Creating single analysis report when multiple dataviews (layers) are selected by the user
- [x] Updated tests to ensure payload is formatted properly

### DIFF
--- a/applications/fishing-map/src/features/analysis/Analysis.tsx
+++ b/applications/fishing-map/src/features/analysis/Analysis.tsx
@@ -163,20 +163,24 @@ function Analysis() {
   }
 
   const onDownloadClick = async () => {
-    const createReports: CreateReport[] = dataviews.map((dataview) => {
+    const reportDataviews = dataviews.map((dataview) => {
       const trackDatasets: Dataset[] = (dataview?.config?.datasets || [])
         .map((id: string) => dataview.datasets?.find((dataset) => dataset.id === id))
         .map((dataset: Dataset) => getRelatedDatasetByType(dataset, DatasetTypes.Tracks))
 
       return {
-        name: `${dataview.name} - ${t('common.report', 'Report')}`,
-        dateRange: staticTime as DateRange,
         filters: dataview.config?.filters || [],
         datasets: trackDatasets.map((dataset: Dataset) => dataset.id),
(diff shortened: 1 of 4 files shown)


applications/fishing-map/src/features/analysis/Analysis.tsx
```diff
@@ -163,20 +163,24 @@ function Analysis() {
...
-        dateRange: staticTime as DateRange,
         filters: dataview.config?.filters || [],
         datasets: trackDatasets.map((dataset: Dataset) => dataset.id),
-        geometry: analysisGeometry as ReportGeometry,
       }
     })
-    await dispatch(createReportThunk(createReports))
+    const reportName = Array.from(new Set(dataviews.map((dataview) => dataview.name))).join(',')
+    const createReport: CreateReport = {
+      name: `${reportName} - ${t('common.report', 'Report')}`,
```
j8seangel, line 178: In spanish we use Report {name} instead of {name} report, so what about using the [interpolation](https://www.i18next.com/translation-function/interpolation) and allow that?
Also the dataview names are not translated and we could use the dataset id as we do for the sidebar.
Another question is how far we want to go into this.. does it really worth it?

applications/fishing-map/src/features/analysis/analysis.slice.ts
```diff
@@ -12,6 +12,7 @@ import {
 } from 'utils/async-slice'
 import { RootState } from 'store'
 import { Bbox } from 'types'
+import { transformFilters } from './analysis.utils'
```
j8seangel, line 15: Thanks for this!

j8seangel (commented):
LGTM, the i18n is a minor issue we can manage later. Thanks!

applications/fishing-map/src/features/analysis/Analysis.tsx
rdgfuentes: The report name is not used in any place yet but it is required to create the report. I'll leave it as it is for now to keep it super simple, I don't think it worths investing more time on it at this point. Makes sense @j8seangel ?

PR #460, opened 2021-04-26
Autoremove unused keys in the translations catalog

rdgfuentes:
- [x] Updated `yarn i18n:parser` config to autoremove unused keys in the code from the catalog (crowdin)

### DIFF
--- a/applications/fishing-map/public/locales/source/translations.json
+++ b/applications/fishing-map/public/locales/source/translations.json
@@ -9,13 +9,7 @@
     "noActivityLayers": "No activity layers active",
     "timeRangeTooLong": "Reports are only allowed for time ranges up to one year",
     "title": "Analysis",
-    "vesselFlags": "by vessels flagged by",
-    "area": "Area",
-    "generating": "We are generating the report for you, it'll be ready soon.",
-    "successfullMessage": "The report was sent successfully, you'll receive it by email soon.",
-    "dateRange": "Date Range",
-    "emailDescription": "A report for the selected date ranges and filters will be generated and sent to your email account",
-    "generate": "Generate"
+    "vesselFlags": "by vessels flagged by"
   },
   "common": {
     "active_after": "Active after",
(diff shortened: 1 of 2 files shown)


applications/fishing-map/public/locales/source/translations.json
```diff
@@ -238,43 +221,6 @@
...
-    "lastUpdate": "Last update",
-    "setBookmark": "Bookmark current time range",
-    "day": "day",
-    "year": "year",
-    "month": "month",
-    "hour": "hour",
-    "zoomIn": "Zoom in",
-    "zoomTo": "Zoom to",
-    "zoomOut": "Zoom out",
-    "selectTimeRange": "Select a time range"
```
rdgfuentes, line 278: I suspect all of these are still in use. What could we do so that i18next-parser does not remove them?

fertandil87: yes I see lot of keys that doend't feel that were removed, did you try some scanner plugins? like
https://github.com/i18next/i18next-scanner/
there are more @rdgfuentes

j8seangel: As the parser only reads `src` code to find keys... there are some use cases where the key isn't in the code but needs to appear in sources, for example:
- API dynamic fields from vessels ([workaround](https://github.com/GlobalFishingWatch/frontend/blob/e7809a4a7977ef62b554f1f29a3a85cad5a86509/applications/fishing-map/src/features/workspace/vessels/VesselTranslations.ts))
- Timebar translations, we don't use the keys in our code but pass the object configuration into the component. How could we manage this ?

Another idea is to update the crowdin sync to update the sources and allow deleting from there too.

Anyway I'd do a round of cleaning in this PR to check which one are false positives and try to keep the source as clean as possible

rdgfuentes: @fertandil87 I tried it briefly but could not make it work with Typescript. Maybe this other one https://www.npmjs.com/package/i18next-scanner-typescript could work but have not tried it.

@j8seangel It's something to think about...  Thanks

j8seangel:
Closing this in favour of https://github.com/GlobalFishingWatch/frontend/pull/474

PR #466, opened 2021-04-28
Layer composer/fix to white

j8seangel:
Make the color ramp white end optional and disable it for environmental layers.

Also fixes:
- [x] Icon crash when no icon found
- [x] Analysis infinite spinner

### DIFF
--- a/packages/dataviews-client/src/resolve-dataviews-generators.ts
+++ b/packages/dataviews-client/src/resolve-dataviews-generators.ts
@@ -125,6 +125,7 @@ export function getGeneratorConfig(
           {
             id: generator.id,
             colorRamp: dataview.config?.colorRamp as ColorRampsIds,
+            colorRampWhiteEnd: false,
             visible: dataview.config?.visible ?? true,
             breaks: dataview.config?.breaks,
             datasets: datasetsIds,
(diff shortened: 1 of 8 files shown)


packages/layer-composer/src/generators/heatmap/util/get-legends.ts
```diff
@@ -6,22 +6,17 @@ import getBreaks from './get-breaks'
 
 // Get color ramps for a config's sublayers
 export const getSublayersColorRamps = (config: GlobalHeatmapAnimatedGeneratorConfig) => {
-  let colorRampIds = config.sublayers.map((s) => s.colorRamp)
-
   // Force bivariate color ramp depending on config
   if (config.mode === HeatmapAnimatedMode.Bivariate) {
-    colorRampIds = ['bivariate']
+    return config.sublayers.map(() => HEATMAP_COLOR_RAMPS['bivariate'])
```
nerik, line 11: I don't understand this. Bivariate mode only uses one color ramp, always.

packages/dataviews-client/src/resolve-dataviews-generators.ts
```diff
@@ -125,6 +125,7 @@ export function getGeneratorConfig(
           {
             id: generator.id,
             colorRamp: dataview.config?.colorRamp as ColorRampsIds,
+            colorRampWhiteEnd: false,
```
nerik, line 128: I'm having doubts on this. It's `fishing-map` application configuration done in package logic. To me this looks like something that should be configured on the dataview, along with more stuff that is done here depending on if environmental layer or nor.

PR #476, opened 2021-05-04
Fishing map/color cycling

nerik:
Adds generic color cycling mechanism when adding a dataview to the map:
- picks a fill or line color palette
- picks the first color of this palette which have the least occurrences amongst all workspace dataviews (visible or not)

### DIFF
--- a/applications/fishing-map/src/features/workspace/workspace.hook.ts
+++ b/applications/fishing-map/src/features/workspace/workspace.hook.ts
@@ -38,14 +38,19 @@ const createDataviewsInstances = (
   return newDataviewInstances.map((dataview) => {
     if (dataview.config?.colorCyclingType) {
       const nextColor = getNextColor(dataview.config.colorCyclingType, currentDataviewInstances)
-      return {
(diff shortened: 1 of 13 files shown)


applications/fishing-map/src/features/workspace/workspace.hook.ts
```diff
@@ -1,12 +1,59 @@
...
+): UrlDataviewInstance[] => {
+  return newDataviewInstances.map((dataview) => {
+    if (dataview.config?.colorCyclingType) {
+      const nextColor = getNextColor(dataview.config.colorCyclingType, currentDataviewInstances)
+      return {
+        ...dataview,
+        config: {
+          ...dataview.config,
+          color: nextColor.value,
+          colorRamp: nextColor.id,
```
j8seangel, line 46: This is going to be in the url and it is only used for fill layers, I'd remove with the `colorCyclingType` too

j8seangel (changes requested):
Thanks for actually removing code! just an small improvement in the comment.

PR #479, opened 2021-05-04
Fishing map/presence choice

j8seangel:
Using <Choice /> component to toggle between fishing and presence layers with new dataviews categories: `fishing` and `presence`

https://user-images.githubusercontent.com/10500650/117045687-1480cf00-ad10-11eb-8b63-4e15d6fd144c.mp4

Also disables presence layer interaction on vessel click.

Blocked by https://github.com/GlobalFishingWatch/api-datasets/pull/25 to remove dataseis mock before merging

Pending:
- [ ] Review performance on toggle as rendering takes some time
- [ ] Enable again the bivariate mode with new UI

### DIFF
--- a/applications/fishing-map/public/locales/en/translations.json
+++ b/applications/fishing-map/public/locales/en/translations.json
@@ -20,7 +20,7 @@
   "common": {
     "active_after": "Active after",
     "active_before": "Active Before",
-    "activity": "Vessel activity",
+    "activity": "Activity",
     "apparentFishing": "Apparent Fishing Effort",
     "confirm": "Confirm",
     "context_area_plural": "Reference Layers",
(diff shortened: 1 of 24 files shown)


applications/fishing-map/public/locales/en/translations.json
```diff
@@ -20,7 +20,7 @@
   "common": {
     "active_after": "Active after",
     "active_before": "Active Before",
-    "activity": "Vessel activity",
+    "activity": "Activity",
```
nerik, line 23: ⚠️ Does not work ⚠️

applications/fishing-map/src/features/dataviews/dataviews.selectors.ts
```diff
@@ -106,6 +107,16 @@ export const selectDataviewInstancesResolved = createSelector(
   }
 )
 
+export const selectDataviewInstancesResolved = createSelector(
```
nerik, line 110: Maybe
```suggestion
export const selectDataviewInstancesByActivityCategory = createSelector(
```

nerik (commented):
Looks good! Careful with length of labels in other languages (set line height at the minimum)

applications/fishing-map/public/locales/en/translations.json
j8seangel: Thanks! updated in crowdin

j8seangel (commented):
That selector is used everywhere and that is why I preserved the name to avoid a lot of changes.

PR #480, opened 2021-05-04
MAP-365 Fix vessels advanced search

rdgfuentes:
- Encode all query values in the vessel advanced search
- Fix filter by active after and active before dates


fertandil87 (approved):

fertandil87:
@rdgfuentes let's deploy RF

PR #487, opened 2021-05-06
VH-101 Add vessel track on vessel profile

rdgfuentes:


applications/vessel-history/src/features/workspace/workspace.slice.ts
```diff
@@ -0,0 +1,59 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
j8seangel, line 1: I'd review what things are used from here and try to cleanup as much as possible because there are some things only from fishing-map features

applications/vessel-history/src/store.ts
```diff
@@ -43,6 +48,7 @@ const store = configureStore({
   middleware: (getDefaultMiddleware) =>
     getDefaultMiddleware(defaultMiddlewareOptions)
       .concat(routerQueryMiddleware)
+      .concat(routerWorkspaceMiddleware)
```
j8seangel, line 51: 🗑️

packages/ui-components/src/modal/Modal.tsx
```diff
@@ -24,7 +27,10 @@ function Modal(props: ModalProps) {
     appSelector = 'root',
     header = true,
     title,
+    className,
```
j8seangel, line 30: Thanks for making this more customisable!

j8seangel (changes requested):
Seems like we have a lot of duplicated code from fishing-map, in one hand this is good as we keep projects sync but in other hand we have some unused code here. Just take a look at some examples at the comments. And maybe is not only code, we might are adding too much complexity here.. for example using workspaces, as we will only have one static workspace but I'm maybe missing something... why is it useful?

Would be also nice to discuss what things worth it to move to another package to simplify the workflow, I'd set this as a priority for the next front-end meeting. WDYT ?

Ah! Toggling the layers I found out there is a weird vertical scroll that disappears when the layers are expanded... but in fact it shouldn't ever appear so worth it a review on the height calculation probably using. `100vh` instead of 100%

https://user-images.githubusercontent.com/10500650/120674402-57130400-c494-11eb-921d-042a170a2bff.mp4

applications/vessel-history/src/features/app/app.selectors.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const selectViewport = createSelector([selectUrlViewport], (urlViewport) => {
+  return {
+    zoom: urlViewport?.zoom || DEFAULT_WORKSPACE.zoom,
+    latitude: urlViewport?.latitude || DEFAULT_WORKSPACE.latitude,
+    longitude: urlViewport?.longitude || DEFAULT_WORKSPACE.longitude,
+  }
+})
+
+export const selectTimeRange = createSelector([selectUrlTimeRange], ({ start, end }) => {
```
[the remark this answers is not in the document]
rdgfuentes: Well, that's something that I left because one of the suggestions from product was that we could implement some sort of timebar to let the users navigate thru time without having to retrieve all the events at once from the api. Makes sense?

applications/vessel-history/src/features/map/map.selectors.ts
```diff
@@ -1,37 +1,84 @@
...
+      start: DEFAULT_WORKSPACE.start,
+      end: DEFAULT_WORKSPACE.end,
+      token: GFWAPI.getToken(),
+    })
+)
+
+type GetGeneratorConfigParams = {
+  dataviews: UrlDataviewInstance[] | undefined
+  resources: ResourcesState
+  rulers: Generators.Ruler[]
```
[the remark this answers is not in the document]
rdgfuentes: I'll remove the others except staticTime. Thanks

applications/vessel-history/src/features/workspace/workspace.middlewares.ts
```diff
@@ -0,0 +1,25 @@
+import { Middleware, Dispatch } from 'redux'
```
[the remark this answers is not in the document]
rdgfuentes: 🧹

applications/vessel-history/src/features/workspace/workspace.slice.ts
rdgfuentes: Done, thanks!

packages/ui-components/src/modal/Modal.tsx
rdgfuentes: ;)

applications/vessel-history/public/locales/source/translations.json
```diff
@@ -13,12 +13,19 @@
     "historyFor": "history for",
     "previous": "previous"
   },
+  "layer": {
+    "toggleVisibility": "Toggle layer visibility"
+  },
+  "map": {
+    "loading": "Map loading",
```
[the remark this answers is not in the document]
rdgfuentes: fixed

PR #497, opened 2021-05-10
Linting/prettier pass

nerik:
Running prettier once and forget about it


j8seangel:
@rdgfuentes @fertandil87 ensure you update your current branches to avoid conflicts, this updates and [fixes](https://github.com/GlobalFishingWatch/frontend/pull/495) a breaking change with prettier so we ran it in the entire project

PR #514, opened 2021-05-13
Fishing map/presence less than one day

nerik:
- removes a hack in Timebar that caused zoom to one day to actually show 23 hrs 🤦
- throw error when trying to set a delta <1 and aggregation mode is not set to average (case of trying to show presence for less than 1 day)
- do not allow zooming more than one day on timebar when presence is enabled
- [WIP] force a minimum of one day time range when toggling to presence


packages/layer-composer/src/generators/heatmap/heatmap-animated.ts
```diff
@@ -208,6 +209,10 @@ class HeatmapAnimatedGenerator {
       finalConfig.interval
     )
 
+    if (timeChunks.deltaInIntervalUnits === 0 && config.aggregationOperation !== AggregationOperation.Avg) {
+      throw new Error('Trying to show less than 1 interval worth of data for this heatmap layer. This could result in showing misleading information when aggregation mode is not set to avg.')
```
j8seangel, line 213: Throwing a new error might cause an ~white~ error boundaries screen... why not use console.warn or error instead ? or is there already a place to catch that error ?

applications/fishing-map/src/features/timebar/Timebar.tsx
```diff
@@ -155,6 +158,8 @@ const TimebarWrapper = () => {
         bookmarkStart={bookmark?.start}
         bookmarkEnd={bookmark?.end}
         bookmarkPlacement="bottom"
+        minimumRange={1}
+        minimumRangeUnit={activityCategory === 'fishing' ? 'hour' : 'day'}
```
j8seangel, line 162: Love this!

applications/fishing-map/src/features/workspace/heatmaps/HeatmapsSection.tsx
```diff
@@ -29,16 +31,27 @@ function HeatmapsSection(): React.ReactElement {
...
   }, [activityCategory])
 
   const onActivityOptionClick = useCallback(
     (activityOption: ChoiceOption) => {
-      dispatchQueryParams({ activityCategory: activityOption.id as WorkspaceActivityCategory })
+      const queryParams: Record<string, any> = {
+        activityCategory: activityOption.id as WorkspaceActivityCategory,
+      }
+      // TODO check that range is < 1 day
+      if (activityOption.id === 'presence' && start) {
```
j8seangel, line 46: once this is done I think we are good to merge, good job!

j8seangel:
We can't merge this yet as there is an issue with the build, it seems to be working on dev but doesn't in production (still investigating)

image

PR #530, opened 2021-05-17
Fishing map/refacto cleanup of staticTime

nerik:
- fixes spamming the 4w API when expanding timeline delta
- deprecates use of staticTime in both app and generator and move time to its own atom

### DIFF
--- a/applications/fishing-map/src/features/app/app.selectors.ts
+++ b/applications/fishing-map/src/features/app/app.selectors.ts
@@ -4,13 +4,11 @@ import { APP_NAME, DEFAULT_WORKSPACE } from 'data/config'
 import {
   selectWorkspace,
   selectWorkspaceState,
-  selectWorkspaceTimeRange,
   selectWorkspaceViewport,
 } from 'features/workspace/workspace.selectors'
 import {
   selectQueryParam,
   selectUrlViewport,
-  selectUrlTimeRange,
   selectLocationCategory,
   selectActivityCategory,
 } from 'routes/routes.selectors'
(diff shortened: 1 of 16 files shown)


applications/fishing-map/src/features/timebar/timebar.hooks.ts
```diff
@@ -1,30 +1,69 @@
...
-  const { dispatchQueryParams } = useLocationConnect()
-  const dispatch = useDispatch()
-  const { start, end } = useSelector(selectTimeRange)
-  // TODO needs to be debounced like viewport
-  const dispatchTimeranges = useCallback(
-    (event: { start: string; end: string; source: string }) => {
-      const range = { start: event.start, end: event.end }
-      if (event.source !== 'ZOOM_OUT_MOVE') {
-        dispatch(setStaticTime(range))
+const selectTimeRange = createSelector(
```
j8seangel, line 19: I'd keep this into the .selector file as it is a selector and having it within a .hook file could be confusing

applications/fishing-map/src/features/timebar/Timebar.tsx
```diff
@@ -128,30 +133,46 @@ const TimebarWrapper = () => {
...
+  if (!start || !end) return null
 
   return (
     <div>
       <TimebarComponent
         enablePlayback={true}
         labels={labels}
-        start={start}
-        end={end}
+        start={internalRange ? internalRange.start : start}
```
j8seangel, line 171: I was worried about having a render without the correct value when the internalRange is null and the start wasn't updated yet but seems to be working well!

j8seangel (commented):
Thanks for simplifying it!

applications/fishing-map/src/features/app/app.selectors.ts
```diff
@@ -161,8 +141,6 @@ export const selectCustomWorkspace = createSelector(
       aoi: undefined,
       dataviews: workspace?.dataviews?.map(({ id }) => id as number),
       viewport,
-      startAt: timerange.start,
```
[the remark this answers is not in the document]
j8seangel: I've restored this, but please confirm you deleted because a reason

applications/fishing-map/src/features/app/app.selectors.ts
```diff
@@ -40,16 +38,6 @@ export const selectViewport = createSelector(
   }
 )
 
-export const selectTimeRange = createSelector(
```
j8seangel, line 43: restored this as well as using DEFAULT_TIME_RANGE was causing rendering and requesting 4wings tiles twice

j8seangel (commented):
I've code review some changes and I'm going to merge and test with the other branch and the latest develop changes

PR #532, opened 2021-05-17
request highlighted workspaces from gfw api to avoid loading spreadhs…

j8seangel:
Saving a lot of bundle sizing loading the spreadsheet library only for the feedback from when needed and using the[ new gfw api endpoint](https://github.com/GlobalFishingWatch/api-datasets/pull/27) `v1/highlighted-workspaces:app` 

List of chunks not loaded by default:
image


nerik:
👍

PR #535, opened 2021-05-18
Fishingmap/last minute improvements

satellitestudiodesign:
- Adds some polyfills 
- Improves scrollbars
- Reduces bundle size 🤞 
- Fix ..toWhite legends on some browsers
- ... some more small stuff


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
j8seangel, line 4: updated husky version, new way of configure it

packages/api-client/package.json
```diff
@@ -28,13 +28,11 @@
...
   "dependencies": {
-    "@globalfishingwatch/api-types": "^4.0.0",
     "file-saver": "^2.0.2"
   },
-  "peerDependencies": {
-    "@globalfishingwatch/pbf": "^1.0.10"
-  },
   "devDependencies": {
+    "@globalfishingwatch/api-types": "^4.0.0",
+    "@globalfishingwatch/pbf": "^1.0.10",
```
j8seangel, line 35: This is not perfect but we need to have the `@globalfishingwatch` dependencies in both `devDependencies` and `peerDependencies` to:
- ensure built is done in order (dev)
- not include the dependency in the bundle (peer)

packages/react-hooks/src/use-map-legend/index.ts
```diff
@@ -1 +1,17 @@
+import {
+  LayerMetadataLegend,
+  LayerMetadataLegendBivariate,
+} from '@globalfishingwatch/layer-composer'
+
+// Duplicated from ui-components map-legend to avoid circular dependencies
```
j8seangel, line 6: Ugly as well, but needed

packages/ui-components/src/base.css
```diff
@@ -65,6 +65,14 @@
   --space-S: 1rem;
   --space-M: 2rem;
   --space-L: 4rem;
+
+  /* SCROLLBARS */
+  --scrollbar-track-color: transparent;
```
j8seangel, line 70: niiiiice 🥇

j8seangel (approved):
Great work! and combined with the gzip in nginx now the bundle download size is only 25% of the original one 🎉

PR #541, opened 2021-05-20
LABELER-109 Fix value array track decoding

rdgfuentes:

### DIFF
--- a/packages/data-transforms/src/track-value-array-to-segments/track-value-array-to-segments.test.ts
+++ b/packages/data-transforms/src/track-value-array-to-segments/track-value-array-to-segments.test.ts
@@ -4,8 +4,7 @@ import { Field } from './types'
 describe('trackValueArrayToSegments', () => {
   test('parses elevation properly', () => {
     const valueArray = [
-      -2147483648, 1, 0, -42342480, 47021440, 1562724078, 10300000, -3943000000, 270600000, 1,
-      782340,
+      -2147483648, 1, 0, -42342480, 47021440, 1562724078, 10300000, -3943, 270600000, 1, 782340,
     ]
     const fields: Field[] = [
       'lonlat' as Field,
@@ -26,8 +25,8 @@ describe('trackValueArrayToSegments', () => {
           speed: 10.3,
           elevation: -3943,
           course: 270.6,
-          night: 0.000001,
-          distance_from_port: 0.78234,
+          night: 1,
+          distance_from_port: 782340,
         },
       ],
     ])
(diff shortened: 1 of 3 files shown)


packages/data-transforms/src/track-value-array-to-segments/track-value-array-to-segments.test.ts
```diff
@@ -0,0 +1,34 @@
+import { trackValueArrayToSegments } from './track-value-array-to-segments'
+import { Field } from './types'
+
+describe('trackValueArrayToSegments', () => {
+  test('parses elevation properly', () => {
+    const valueArray = [
+      -2147483648, 1, 0, -42342480, 47021440, 1562724078, 10300000, -3943, 270600000, 1, 782340,
```
fertandil87, line 7: why you are not using DEFAULT_NULL_VALUE here?

packages/data-transforms/src/track-value-array-to-segments/track-value-array-to-segments.ts
```diff
@@ -3,12 +3,36 @@ import { Field, Segment, Point } from './types'
...
+  [Field.course]: (value: number) => value / Math.pow(10, 6),
+  [Field.distanceFromPort]: (value: number) => value,
+  // Uncomment when added to new package version
+  // [Field.distanceFromShore]: (value: number) => value,
+  [Field.elevation]: (value: number) => value,
+  // Uncomment when added to new package version
+  // [Field.encounter]: (value: number) => value,
+  [Field.latitude]: (value: number) => value / Math.pow(10, 6),
+  [Field.longitude]: (value: number) => value / Math.pow(10, 6),
+  [Field.fishing]: (value: number) => value,
```
fertandil87, line 24: I think this is wrong, fishing is a value from 0 to 1 with decimals

fertandil87 (changes requested):

packages/data-transforms/src/track-value-array-to-segments/track-value-array-to-segments.ts
rdgfuentes: No, it is correct. Fishing is an integer, 0 or 1 without decimals. At least that's how it's encoded (https://github.com/GlobalFishingWatch/api-vessels-service/blob/develop/src/modules/api/v1/vessels/modules/tracks/services/feature.service.ts#L17-L24)

packages/data-transforms/src/track-value-array-to-segments/track-value-array-to-segments.test.ts
rdgfuentes: Because here I'm mocking the response from the api endpoint in valueArray format. We shouldn't use it here since we don't know what we're going to receive (in theory).

fertandil87 (approved):

PR #548, opened 2021-05-21
MR-257 Google Analytics

rdgfuentes:


applications/fishing-map/src/features/analysis/Analysis.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useMemo, useState, Fragment } from 'react'
 import cx from 'classnames'
+import * as ua from 'react-ga'
```
j8seangel, line 3: Is there any specific reason to import the entire package ? We might can see some bundle size if importing just what we need.

applications/fishing-map/src/features/workspace/heatmaps/HeatmapsSection.tsx
```diff
@@ -133,3 +165,2129 @@ function HeatmapsSection(): React.ReactElement {
 }
 
 export default HeatmapsSection
+
+// const dv = {
```
j8seangel, line 169: oops 👀

applications/fishing-map/src/features/app/analytics.hooks.ts
```diff
@@ -1,40 +1,49 @@
...
+      //   ua.set({ sendHitTask: null })
+      // }
+    }
+  }, [])
+
+  useEffect(() => {
+    if (userData && GOOGLE_UNIVERSAL_ANALYTICS_ID) {
+      ua.set({
+        user_properties: {
+          user_id: userData.id,
```
j8seangel, line 36: I've just realised this is using snake_case here but camelCase above... is this correct?

j8seangel (commented):
Just a few comments and something large to delete but in general it LGTM

applications/fishing-map/src/features/analysis/Analysis.tsx
rdgfuentes: Ok, I'll change it to import just the methods that are used

applications/fishing-map/src/features/app/analytics.hooks.ts
rdgfuentes: Just to continue setting them as they were when we used gtag... I'll change it to camelCase

applications/fishing-map/src/features/workspace/heatmaps/HeatmapsSection.tsx
rdgfuentes: Fixed

PR #562, opened 2021-05-24
request buckets for zoom 0 and 3

j8seangel:
Trying to get better buckets resolution using 3 as zoom level


j8seangel:
Not needed anymore as dynamic-ramps will solve this better

PR #564, opened 2021-05-25
Update header menu

fertandil87:
Updated the items of the menu
NOT support third level menu


packages/ui-components/src/header/Header.links.ts
```diff
@@ -0,0 +1,101 @@
+export type MenuItem = {
+  link: string
+  label: string
+  mini?: boolean
+  childs?: MenuItem[]
+}
+
+const navigation: MenuItem[] = [
+  {
+    link: 'https://globalfishingwatch.org/topics/',
```
rdgfuentes, line 10: Wouldn't be better to use relative urls for the links so that when this is deployed in any instance is navigable? WDYT @fertandil87 @j8seangel ?

.changeset/tough-vans-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@globalfishingwatch/ui-components": patch
```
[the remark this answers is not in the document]
j8seangel: Indeed! for me this is a breaking change however would be nice to get it automatically updated in our products I think is a better practice to do it by hand and use the versioning

packages/ui-components/src/header/Header.links.ts
j8seangel: I don't think so in this case because some of them could be out of the base url and that url should be permanent

rdgfuentes: Makes sense 👍

.changeset/tough-vans-sneeze.md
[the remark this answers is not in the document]
fertandil87: They are just links changing, don't break anyting. Also is needed the update, maybe some of the old links will break

PR #599, opened 2021-05-28
Fishing map/4wings filter precision

nerik:
This is a fix for https://globalfishingwatch.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=MR&modal=detail&selectedIssue=MR-314

Here's the problem as I explained it to Raul:
Tile 8/68/122 con interval 10days, celda 3534, el tile tiene:
`[..., 3534,1815,1815,100, ...]`
O sea: un solo valor en la frame 1815.
El start del timebar es 17 de septiembre.
Yo calculo hago una conversion de la fecha en frame para saber si mostrar la frame o no, asi:
`Math.floor(timestampMs / 1000 / 60 / 60 / 24 / 10)`
Lo que me devuelve exactemente 1815, con lo cual la celda se puede ver en el mapa, pero al llamar la API se manda el start **del timebar** (17 de septiembre), no **el start del periodo de 10 dias que corresponde a la frame 1815**, es decir 11 de septiembre.

This PR implements a fix suggested by him:
When calling the interaction API, the "real" start and end will be used, meaning start and end frames get converted back to dates, rather than using dates from the timebar (so 11 sept, not 17 sept).

TODO:
- [ ] Since we only do one API call, so only one timerange, this won't work with layers that use different chunks/intervals (ie if we had interaction on env layers with `month` interval). This is an hypothetical scenario but I believe `fetch4winginteractionThunk` should be refactored to use a single set of parameters with an array of datasets.
- [ ] This also raises a design issue:
Screenshot 2021-05-28 at 14 49 23
The range displayed is actually incorrect for the heatmap (but correct for tracks, etc)
   - [ ] We could force the timebar to stick to the 10 days intervals, but it is weird in terms of UI and will conflict with already existing behaviour of sticking to calendar months
   - [ ] We could display a second ramge
   - [ ] Any other suggestion that actually doesn't suck

### DIFF
--- a/applications/fishing-map/src/features/map/map.slice.ts
+++ b/applications/fishing-map/src/features/map/map.slice.ts
@@ -91,6 +91,7 @@ type SublayerVessels = {
 
 export const fetch4WingInteractionThunk = createAsyncThunk<
   { vessels: SublayerVessels[] } | undefined,
+  // TODO the whole function could be greatly simplified if only one temporalGridFeature was accepted, which is effecttively always the case
   { temporalGridFeatures: ExtendedFeature[]; timeRange: Range },
   {
     dispatch: AppDispatch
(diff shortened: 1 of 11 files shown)


.changeset/curvy-buttons-brake.md
```diff
@@ -0,0 +1,8 @@
+---
+"@globalfishingwatch/fourwings-aggregate": patch
+"@globalfishingwatch/layer-composer": patch
+"@globalfishingwatch/react-hooks": patch
+"@globalfishingwatchapp/fishing-map": major
```
nerik, line 5: Let's move to 1.0.0 !!

applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -91,6 +91,7 @@ type SublayerVessels = {
 
 export const fetch4WingInteractionThunk = createAsyncThunk<
   { vessels: SublayerVessels[] } | undefined,
+  // TODO the whole function could be greatly simplified if only one temporalGridFeature was accepted, which is effecttively always the case
```
nerik, line 94: What I meant was that we only need an array of datasets, all other parameters have to be the same

j8seangel:
Showing the real timeRange for the 10days interval in the tooltip now seems to be a good workaround for now:
image

packages/react-hooks/src/use-map-interaction/index.ts
```diff
@@ -1,9 +1,21 @@
 import type { Geometry } from 'geojson'
 import { ContextLayerType } from '@globalfishingwatch/layer-composer/dist/generators/types'
+import { Interval } from '@globalfishingwatch/layer-composer/dist/generators'
 import { PointerEvent } from '@globalfishingwatch/react-map-gl'
 
 export { useMapHover, useMapClick, useFeatureState } from './use-map-interaction'
 
+export type TemporalGridFeature = {
```
nerik, line 8: YES ! 👍

PR #618, opened 2021-06-15
VE-74 Retrieve eezs and rfmos from config endpoint

rdgfuentes:


rdgfuentes:
Continued on #627 with a different approach

PR #620, opened 2021-06-16
fixed issue with 4w time range + more greedy

nerik:
Screenshot 2021-06-16 at 17 54 04

- fixes an issue where the last frame was missing in both map and popup
- delta is now calculated using `ceil` for time range end, which can result in another extra frame queried

### DIFF
--- a/applications/fishing-map/src/features/map/map.slice.ts
+++ b/applications/fishing-map/src/features/map/map.slice.ts
@@ -114,14 +114,8 @@ export const fetch4WingInteractionThunk = createAsyncThunk<
 
     // Currently only one timerange is supported, which is OK since we only need interaction on the activity heatmaps and all
     // activity heatmaps use the same time intervals, This will need to be revised in case we support interactivity on environment layers
-    const start =
-      mainFeature.temporalgrid?.interval === '10days'
-        ? mainFeature.temporalgrid?.visibleStartFrame
-        : timeRange.start
-    const end =
-      mainFeature.temporalgrid?.interval === '10days'
-        ? mainFeature.temporalgrid?.visibleEndFrame
-        : timeRange.end
+    const start = mainFeature.temporalgrid?.visibleStartDate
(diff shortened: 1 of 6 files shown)


applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -116,11 +116,11 @@ export const fetch4WingInteractionThunk = createAsyncThunk<
     // activity heatmaps use the same time intervals, This will need to be revised in case we support interactivity on environment layers
     const start =
       mainFeature.temporalgrid?.interval === '10days'
```
nerik, line 118: I actually think this condition should be removed, and vessel queries should always be made depending on the real 4w time range, not the one in timebar (even though in practice it only makes a difference for `10days` because of timebar 'stickiness' with day/hour)

j8seangel (commented):
Just not an important comment of naming, but please confirm the changed I did to get it working. The tooltip now shows the expected time range 👍 
image

applications/fishing-map/src/features/map/map.slice.ts
j8seangel: We used to have a bug with the `day` period where the value of `visibleStartFrame` didn't match the real one. But it seems to be fixed with these changes. Do you want me to do it?

packages/react-hooks/src/use-map-interaction/index.ts
```diff
@@ -12,8 +12,8 @@ export type TemporalGridFeature = {
   col: number
   row: number
   interval: Interval
-  visibleFramesStart: string
-  visibleFramesEnd: string
+  visibleStartFrame: string
```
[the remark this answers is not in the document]
nerik: Good catch, thanks

applications/fishing-map/src/features/map/map.slice.ts
nerik: Did it, it's working fine

PR #633, opened 2021-06-25
improved descriptions of the events

fertandil87:
added better description to the events of the activity

translations still missing until I have more event types an a better understanding of how the string will be built

image

### DIFF
--- a/applications/vessel-history/src/features/profile/components/activity/Activity.tsx
+++ b/applications/vessel-history/src/features/profile/components/activity/Activity.tsx
@@ -14,6 +14,7 @@ import { fetchRegionsThunk } from 'features/regions/regions.slice'
 import styles from './Activity.module.css'
 import ActivityDate from './ActivityDate'
 import ActivityDescription from './description/ActivityDescription'
+import ActivityGroupDescription from './description/ActivityGroupDescription'
 
 interface InfoProps {
   vessel: VesselWithHistory | null
@@ -58,11 +59,11 @@ const Activity: React.FC<InfoProps> = (props): React.ReactElement => {
                     </div>
                     <div className={styles.eventData}>
                       <div className={styles.date}>
-                        <I18nDate date={group.entries[0].start} format={DateTime.DATE_SHORT} /> - 
-                        <I18nDate date={group.entries[group.entries.length - 1].end} format={DateTime.DATE_SHORT} />
+                        <I18nDate date={group.start} format={DateTime.DATE_SHORT} /> - 
+                        <I18nDate date={group.end} format={DateTime.DATE_SHORT} />
                       </div>
                       <div className={styles.description}>
-                        <ActivityDescription  group={group} ></ActivityDescription>
+                        <ActivityGroupDescription  group={group} ></ActivityGroupDescription>
                       </div>
                       
                     </div>
--- a/applications/vessel-history/src/features/profile/components/activity/Activity.tsx
+++ b/applications/vessel-history/src/features/profile/components/activity/Activity.tsx
@@ -10,8 +10,10 @@ import { fetchVesselActivityThunk } from 'features/vessels/activity/vessels-acti
 import { selectVesselActivity, toggleGroup } from 'features/vessels/activity/vessels-activity.slice'
 import { ActivityEvent, ActivityEventGroup } from 'types/activity'
 import I18nDate from 'features/i18n/i18nDate'
+import { fetchRegionsThunk } from 'features/regions/regions.slice'
 import styles from './Activity.module.css'
 import ActivityDate from './ActivityDate'
+import ActivityDescription from './description/ActivityDescription'
 
 interface InfoProps {
   vessel: VesselWithHistory | null
@@ -37,6 +39,10 @@ const Activity: React.FC<InfoProps> = (props): React.ReactElement => {
     dispatch(toggleGroup({index}))
   },[dispatch])
 
+  useEffect(() => {
+    dispatch(fetchRegionsThunk())
+  }, [dispatch])
+  
   const eventGroups = useSelector(selectVesselActivity)
   return (
     <Fragment>
@@ -56,7 +62,7 @@ const Activity: React.FC<InfoProps> = (props): React.ReactElement => {
                         <I18nDate date={group.entries[group.entries.length - 1].end} format={DateTime.DATE_SHORT} />
                       </div>
                       <div className={styles.description}>
-                      {group.entries.length} Fishing events in {group.event_place}
+                        <ActivityDescription  group={group} ></ActivityDescription>
                       </div>
                       
                     </div>
@@ -78,7 +84,7 @@ const Activity: React.FC<InfoProps> = (props): React.ReactElement => {
                     <div className={styles.eventData}>
                       <ActivityDate event={event}/>
                       <div className={styles.description}>
-                        Fishing in ???? {event.start}
+                        <ActivityDescription  event={event} ></ActivityDescription>
                       </div>
                       
                     </div>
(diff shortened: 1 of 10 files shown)


applications/vessel-history/src/features/regions/regions.selectors.ts
```diff
@@ -19,3 +19,21 @@ export const selectMPAs = selectRegionsById(MarineRegionType.mpa)
...
+    return eezs.find(eez => eez.id == id)
+  })
+)
+
+export const selectRfmoById = memoize((id: RegionId) =>
+  createSelector([selectRFMOs], (eezs) => {
+    if (!id) {
+      return null
+    }
+    return eezs.find(eez => eez.id == id)
```
rdgfuentes, line 37: Use `rfmos` for the variable name in the selector instead of `eezs`

applications/vessel-history/src/features/profile/components/activity/description/ActivityDescription.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return <ActivityDescriptionEEZ regionId={event.regions.eez[0]} type="event" ocean={event.regions.ocean[0]}></ActivityDescriptionEEZ>
+  }
+
+  if (group?.event_rfmo){
+    return <ActivityDescriptionEEZ regionId={group?.event_rfmo} type="group" ocean={group.ocean}></ActivityDescriptionEEZ>
+  }
+
+  if (event?.regions.rfmo[0]){
+    return <ActivityDescriptionEEZ regionId={event.regions.rfmo[0]} type="event" ocean={event.regions.ocean[0]}></ActivityDescriptionEEZ>
+  }
```
rdgfuentes, line 28: There are too many `if` conditions in this component, this could probably be symptoms that something should be improved in the way we are structuring the data or the purpose/responsibility of this component.

applications/vessel-history/src/features/profile/components/activity/description/ActivityDescriptionEEZ.tsx
```diff
@@ -0,0 +1,41 @@
...
+  ocean?: string
+  count?: number
+}
+
+const ActivityDescriptionEEZ: React.FC<ActivityDescriptionProps> = (props): React.ReactElement => {
+  const regionId = props.regionId
+  const { t } = useTranslation()
+
+  const region = useSelector(selectEezById(regionId))
+  const regionName = region?.label?.replace('Exclusive Economic Zone', 'EEZ') || props.ocean || 'unknown'
```
rdgfuentes, line 19: If we need to modify/normalize the labels of the regions I'd try to do that in other place and not in a component before rendering, because it'll only be applied for this component. Instead I'll try to apply those transformations before storing the regions data in the state (could be the thunk) so that's already fixed there and available for the whole app. Makes sense?

applications/vessel-history/src/features/profile/components/activity/Activity.tsx
```diff
@@ -37,6 +39,10 @@ const Activity: React.FC<InfoProps> = (props): React.ReactElement => {
     dispatch(toggleGroup({index}))
   },[dispatch])
 
+  useEffect(() => {
+    dispatch(fetchRegionsThunk())
```
rdgfuentes, line 43: If we're going to call `fetchRegionsThunk` every time the user enters a vessel profile, I'd add a condition to the thunk so that it performs the request to the API only once.

applications/vessel-history/src/features/profile/components/activity/Activity.tsx
```diff
@@ -56,7 +62,7 @@ const Activity: React.FC<InfoProps> = (props): React.ReactElement => {
                         <I18nDate date={group.entries[group.entries.length - 1].end} format={DateTime.DATE_SHORT} />
```
rdgfuentes, line 62: Start and end time of the group should be set in the state too

applications/vessel-history/src/types/activity.ts
```diff
@@ -55,7 +55,9 @@ export interface ActivityEvent {
 }
 export interface ActivityEventGroup {
     event_type: EventType
-    event_place?: string
+    event_eez?: number | string
+    event_rfmo?: number | string
+    ocean?: string
```
rdgfuentes, line 60: I'd try to do this more generic to include any region and not having to explicitly modify this type  in the future.

applications/vessel-history/src/features/vessels/activity/vessels-activity.thunk.ts
```diff
@@ -44,15 +44,26 @@ const groupEvents = (events: ActivityEvent[]) => {
...
+                ocean: event.regions.ocean[0],
                 open: true,
                 entries: [event]
             })
-        } else if (groups[groups.length - 1].event_type === event.type && groups[groups.length - 1].event_place === event.regions.eez[0]) {
+        } else if (
+            groups[groups.length - 1].event_type === event.type &&
+            (
+                groups[groups.length - 1].event_eez === event.regions.eez[0] &&
+                groups[groups.length - 1].event_rfmo === event.regions.rfmo[0]
```
rdgfuentes, line 64: What about the `ocean`? shouldn't it be verified too?

applications/vessel-history/src/features/profile/components/activity/Activity.tsx
fertandil87: not understand this

applications/vessel-history/src/features/vessels/activity/vessels-activity.thunk.ts
```diff
@@ -44,15 +44,26 @@ const groupEvents = (events: ActivityEvent[]) => {
...
+                groups[groups.length - 1].event_eez !== event.regions.eez[0] ||
+                groups[groups.length - 1].event_rfmo !== event.regions.rfmo[0]
+            )
         ) {
             groups.push({
                 event_type: event.type,
-                event_place: event.regions.eez[0],
+                event_eez: event.regions.eez[0],
+                event_rfmo: event.regions.rfmo[0],
+                ocean: event.regions.ocean[0],
```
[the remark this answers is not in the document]
fertandil87: because at the moment I did this, I asked for the criteria to create the grops but is not defined, so did ot based on what I can see in the design and the data, maybe this can change in the future or they like my solution

applications/vessel-history/src/features/vessels/activity/vessels-activity.thunk.ts
fertandil87: because at the moment I did this, I asked for the criteria to create the grops but is not defined, so did ot based on what I can see in the design and the data, maybe this can change in the future or they like my solution

applications/vessel-history/src/features/profile/components/activity/description/ActivityDescriptionEEZ.tsx
fertandil87: maybe yes but I am not sure if this replace is needed, I did it to fit with the design

applications/vessel-history/src/features/profile/components/activity/description/ActivityDescription.tsx
fertandil87: well this component is the one that is doing and improvement, verifing what data we have and calling the right descriptions so the selectors with ID are called only when the right id is present and is not possible to have the selector in a IF

applications/vessel-history/src/types/activity.ts
fertandil87: I can add an array but will add lot logic to make build the groups, and render data. maybe make sense, I will take a look

PR #638, opened 2021-06-29
fetch viirs interaction

j8seangel:
Fetch VIIRS cell stats thought the 4wings interaction endpoint to render:

image

to confirm with @satellitestudiodesign the styles of the modal


satellitestudiodesign:
I've tweaked the styles to make the popup consistent with the fishing effort one.

About **interaction endpoint payload is unnecessarily big**: it certainly is right now, same with the fishing effort one, but keeping them disaggregated like this we'll enable us to create the next iteration of these popups, where we can show things like:
- a bar chart of the distribution of vessels by flag
- a histogram of the radiance of viirs detections

About **enabling presence and VIIRS layers at the same time, we are comparing things with distinct units**: True. apart from the amount of false positives, we cannot call VIIRS unit "days" because satellites pass over an area twice a day so the comparisons are broken. We can remove the presence layer in the VIIRS workspace for now, but we have to at least start thinking about the proper way to solve this.

PR #641, opened 2021-06-30
fetch viirs interaction technical refactos

nerik:
- merged code to prepare fishing and viirs interaction endpoints, only viirs is an array of one
- pass `dataset.config.type` to sublayer metadata in order to determine type of endpoint to call, rather than using`temporalgrid.sublayerId.startsWith` which seem unreliable to me
- some more fluff

### DIFF
--- a/applications/fishing-map/src/features/map/map.hooks.ts
+++ b/applications/fishing-map/src/features/map/map.hooks.ts
@@ -141,7 +141,8 @@ export const useClickedEventConnect = () => {
     }
 
     // Cancel all pending promises
-    [fishingPromiseRef, viirsPromiseRef, eventsPromiseRef].forEach(ref => {
+    const promisesRef = [fishingPromiseRef, viirsPromiseRef, eventsPromiseRef]
+    promisesRef.forEach((ref) => {
       if (ref.current) {
         ref.current.abort()
       }
@@ -187,8 +188,8 @@ export const useClickedEventConnect = () => {
         return false
       }
       const isFeatureVisible = feature.temporalgrid.visible
-      const isFishingFeature = feature.temporalgrid.sublayerInteractionType === 'viirs'
-      return isFeatureVisible && isFishingFeature
+      const isViirsFeature = feature.temporalgrid.sublayerInteractionType === 'viirs'
+      return isFeatureVisible && isViirsFeature
     })
 
     if (viirsFeature) {
--- a/applications/fishing-map/src/features/map/map.hooks.ts
+++ b/applications/fishing-map/src/features/map/map.hooks.ts
@@ -176,14 +176,14 @@ export const useClickedEventConnect = () => {
           return false
         }
         const isFeatureVisible = feature.temporalgrid.visible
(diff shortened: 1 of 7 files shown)


applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -281,62 +286,17 @@ export const fetchFishingActivityInteractionThunk = createAsyncThunk<
...
-      },
-    ],
-  }
-
-  const filters = featureDataview.config?.filter
-  if (filters) {
-    const filtersArray = Array.isArray(filters) ? filters : [filters]
-    datasetConfig.query?.push({ id: 'filters', value: filtersArray })
-  }
+  const { fourWingsDataset, datasetConfig } = getInteractionEndpointDatasetConfig(getState, [feature])
```
nerik, line 299: Basically just this

applications/fishing-map/src/features/map/map.hooks.ts
```diff
@@ -192,7 +187,7 @@ export const useClickedEventConnect = () => {
         return false
       }
       const isFeatureVisible = feature.temporalgrid.visible
-      const isFishingFeature = feature.temporalgrid.sublayerId.startsWith('viirs')
+      const isFishingFeature = feature.temporalgrid.sublayerInteractionType === 'viirs'
       return isFeatureVisible && isFishingFeature
```
j8seangel: copied and pasted from above... and I forgot to rename to `isViirsFeature` just fixed

applications/fishing-map/src/features/map/map.slice.ts
```diff
@@ -95,73 +95,78 @@ type SublayerVessels = {
   vessels: ExtendedFeatureVessel[]
 }
 
+const getInteractionEndpointDatasetConfig = (getState: () => any, features: ExtendedFeature[]) => {
+  const state = getState() as RootState
```
j8seangel: In this case where only the temporalgridDataviews param is need, for me it is easier to have just the param instead of the function which returns the state... so I replaced the getState with:
```js
const getInteractionEndpointDatasetConfig = (
  features: ExtendedFeature[],
  temporalgridDataviews: UrlDataviewInstance[] = []
)
```
also to take advantage of typings

applications/fishing-map/src/features/map/map.slice.ts
j8seangel: Thanks! it is great

packages/layer-composer/src/generators/types.ts
```diff
@@ -356,6 +356,8 @@ export type Ruler = {
   isNew?: boolean
 }
 
+export type HeatmapAnimatedInteractionType = 'presence' | 'viirs' | 'fishing-effort'
```
j8seangel, line 359: This is way more obvious than the previous way of identifying the interaction 🙇

j8seangel (commented):
I love it! thanks for simplifying and making it clearer. I've directly committed the CR and the changesets so this is ready to merge.

PR #642, opened 2021-07-01
Vessel Activity to reuse MAP approach

rdgfuentes:
- [x] Move resolve resource queries to dataviews-client
- [x] Move resource reducer and selectors to dataviews-client
- [x] Build vessel activity (fishing and loitering) from vessel resource
- [x] Remove vessel activity slice
- [x] Display loitering events in vessel activity tab
- [x] Display loitering event details when user clicks on `i`


applications/fishing-map/src/features/resources/resources.selectors.ts
```diff
@@ -28,51 +20,7 @@ export const selectVisibleResources = createSelector(
...
-
-      return [trackResource, infoResource, ...eventsResources].filter(
-        (r) => r !== null
-      ) as Resource[]
-    })
-
-    return resourceQueries
-  }
+  [selectDataviewInstancesResolved],
+  (dataviewInstances) => resolveResourceEndpoint(dataviewInstances ?? [])
```
j8seangel, line 25: I love how clean this is now 👏 but the method name looks confusing, doesn't it?
It receives dataviews and returns queries, so why not calling it something like `resolveDataviewsQueries` ?

applications/fishing-map/src/features/resources/resources.slice.ts
```diff
@@ -1,93 +1,13 @@
...
-import GFWAPI from '@globalfishingwatch/api-client'
 import {
-  Resource,
-  ResourceStatus,
-  DatasetTypes,
-  ApiEvent,
-  ApiEvents,
-} from '@globalfishingwatch/api-types'
-import { RootState } from 'store'
+  ResourcesState as CommonResourcesState,
```
j8seangel, line 2: 😍

packages/dataviews-client/src/resources/resources-endpoint.ts
```diff
@@ -0,0 +1,55 @@
+import { DatasetTypes, Resource } from '@globalfishingwatch/api-types'
```
j8seangel, line 1: The same comment about naming, if we change the method name I'd change the file name too

packages/dataviews-client/src/resources/resources-slice.ts
```diff
@@ -0,0 +1,100 @@
...
+      state[url].status = ResourceStatus.Error
+    })
+  },
+})
+
+export const selectResources = (state: PartialStoreResources) => state.resources
+export const selectResourceByUrl = memoize(<T = any>(url = '') =>
+  createSelector([selectResources], (resources) => resources[url] as Resource<T>)
+)
+export const selectResourcesLoading = createSelector([selectResources], (resources) => {
```
j8seangel, line 94: nice and useful to have it available for other apps

j8seangel (commented):
Great work! apart from that small comment it looks great thanks for simplifying things 🙇 and glad you could reuse "easily" some parts of the fishing map

applications/fishing-map/src/features/resources/resources.selectors.ts
rdgfuentes: Thanks! Yes, you are right, I'll rename it to `resolveDataviewsResourceQueries` to express that the queries being resolved belong to resources.

PR #655, opened 2021-07-06
Fishing map/vessel events perf

nerik:
- removed highlightedTime as timebar wrapper dependency, in order to avoid rerendering timebar subcharts when hovering
- timebar children are now rendered directly as a react node, not a callback
- track events are filtered by visible outer delta
- tracxk events are clustered for better perf when viewing large span of time


applications/fishing-map/src/features/timebar/Timebar.tsx
```diff
@@ -83,9 +96,7 @@ const TimebarWrapper = () => {
     (clientX: number, scale: (arg: number) => Date) => {
       if (hoverInEvent.current === false) {
         if (clientX === null || clientX === undefined || isNaN(clientX)) {
-          if (highlightedTime !== undefined) {
-            dispatch(disableHighlightedTime())
-          }
+          dispatch(disableHighlightedTime())
```
j8seangel, line 99: I think this was necessary to avoid spamming the store, did you check it?

packages/timebar/src/charts/tracks-events.js
```diff
@@ -95,13 +95,33 @@ const getCoordinates = (tracksEvents, outerScale) => {
...
+ */
+const getFilteredEvents = (tracksEvents, outerStart, outerEnd) => {
+  const outerStartMs = new Date(outerStart).getTime()
+  const outerEndMs = new Date(outerEnd).getTime()
+  return tracksEvents.map(trackEvents => {
+    const filtered = trackEvents.filter(event => {
+      return event.end > outerStartMs && event.start < outerEndMs
+    })
+    
+    console.log(trackEvents.length, filtered.length)
```
j8seangel, line 109: 👀

j8seangel (commented):
Very good improvements and code cleaning! please take a look at the small comments and don't forget to release a major version of the timebar

nerik:
Here's how clustering of events work:

https://user-images.githubusercontent.com/1583415/124576502-a8743180-de4c-11eb-888a-f03a23e806e6.mov

- no clustering happens below 2 months of visible time
- tooltip changes to a more generic message when hovering over a clustered event (will need some i18n trickery)
- when hovering over a clustered event, map callback is not called to avoid wrongly highlighting only one event in the map. One possible improvement could be to highlight all clustered events.

applications/fishing-map/src/features/timebar/Timebar.tsx
nerik: It will set an already `undefined` value to `undefined`, so it shouldn't trigger any change on the store, correct?
How can we be 100% sure of this. This was needed to avoid having TimebarWrapper having a dependency on a 'high-frequency' prop.

PR #669, opened 2021-07-08
Fix perf issue on track events hover

nerik:
Fix for Enrique's comment:
> one question about fishing event performance, I'm looking into this workspace. When I'm at the zoom level where the fishing events become part of the track the map became really unresponsive to me

This is due to merging all events datasets in https://github.com/GlobalFishingWatch/frontend/compare/fishing-map/fix-perf-issue-on-track-events-hover?expand=1#diff-f9708754c8ace51fecd0d6922da43e4dfa2044b5729a1f605a99b98043104df7R130, which generates a new array for events every time generators are reinstanciated. Fix is a semi-hack where the generator uses a custome memoizeOne equality check.

This is a bit unsatisfying and reveals larger underlying issues that we should look at fixing post-release:
- `getVesselEventsSegmentsGeojson` is very slow, around two seconds - will need to be optimized
- we should start thinking of a mechanism so that the entire `resolve dataviews -> instanciate generators -> run generators` cycle does not run. 
   - at every highlighted time update
   - at every current frame update
   The `memoizeById` stuff was good enough until now, but it happens too late in the process,  and it overcomplicates generators. And, stuff happening in `dataviews-client` and app selectors will get slower and slower.
   
As a side note: do we still use `highlightedEvent` in the track generator? https://github.com/GlobalFishingWatch/frontend/pull/669/files#diff-d426994ad6a0cd7c1064563fb7860319bee5408c0c333fd33a4b8be62f8adc15R132

### DIFF
--- a/.changeset/stale-hounds-search.md
+++ b/.changeset/stale-hounds-search.md
@@ -1,0 +1,6 @@
+---
+"@globalfishingwatch/dataviews-client": patch
+"@globalfishingwatch/layer-composer": patch
+---
+
+Fix perf issue on track events hover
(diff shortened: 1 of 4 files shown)


j8seangel (approved):
Good catch! and I also can't agree more on:
`This is a bit unsatisfying and reveals larger underlying issues that we should look at fixing post-release:` lets merge and add it to the TODO list

The highlightedEvent is not useful for fishing-map anymore, but it is for CVP and maybe for vessel-examiner so I'd leave as it is for now

PR #675, opened 2021-07-09
Fixes popup units

nerik:
Allows specifying units by sublayer, not using the same for the whole AnimatedHeatmap.
This is the only interesting thing https://github.com/GlobalFishingWatch/frontend/pull/675/files#diff-da964bda90275d1e502b46a7f6f87331584e044ed5ed6844731a05bc6dec8b7aR104

Fixes MAP-443 https://globalfishingwatch.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=MAP-443&assignee=5f1f29eea1648700150d0631&assignee=5f1f2a4ecdb7b4001bb074e5

### DIFF
--- a/applications/fishing-map/src/features/map/popups/EnvironmentLayers.tsx
+++ b/applications/fishing-map/src/features/map/popups/EnvironmentLayers.tsx
@@ -32,6 +32,7 @@ function EnvironmentTooltipSection({
               <div className={styles.row}>
                 <span className={styles.rowText}>
                   {parseEnvironmentalValue(feature.value)}{' '}
+                  {/* TODO will need to not pick from temporalgrid once user polygons support units  */}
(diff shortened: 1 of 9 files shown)


applications/fishing-map/src/features/map/popups/EnvironmentLayers.tsx
```diff
@@ -32,7 +32,7 @@ function EnvironmentTooltipSection({
               <div className={styles.row}>
                 <span className={styles.rowText}>
                   {parseEnvironmentalValue(feature.value)}{' '}
-                  {feature.unit && <span>{feature.unit}</span>}
+                  {feature.temporalgrid?.unit && <span>{feature.temporalgrid?.unit}</span>}
```
j8seangel, line 35: There are environmental layers which doesn't contain the temporalgrid property but they have unit, so I'd use feature.unit as a fallback of temporalgrid?.unit

PR #685, opened 2021-07-12
disable interaction for attribution links

j8seangel:
https://user-images.githubusercontent.com/10500650/125298277-4887fa00-e328-11eb-8900-6c2bc1544fe5.mp4

### DIFF
--- a/packages/react-hooks/src/use-map-interaction/use-map-interaction.ts
+++ b/packages/react-hooks/src/use-map-interaction/use-map-interaction.ts
@@ -236,15 +236,20 @@ export const useMapHover = (
 
   const onMapHover = useCallback(
     (event) => {
-      // Turn all sources with active feature states off
-      cleanFeatureState()
       const hoverEvent: InteractionEvent = {
         type: 'hover',
         point: event.point,
         longitude: event.lngLat[0],
         latitude: event.lngLat[1],
       }
-      if (event.features?.length) {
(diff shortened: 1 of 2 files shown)


packages/react-hooks/src/use-map-interaction/use-map-interaction.ts
```diff
@@ -236,15 +236,20 @@ export const useMapHover = (
...
-      // Turn all sources with active feature states off
-      cleanFeatureState()
       const hoverEvent: InteractionEvent = {
         type: 'hover',
         point: event.point,
         longitude: event.lngLat[0],
         latitude: event.lngLat[1],
       }
-      if (event.features?.length) {
+      const isLinkHover = event.target.tagName.toLowerCase() === 'a'
```
nerik, line 245:
```suggestion
      const isNotCanvas = event.target.tagName.toLowerCase() !== 'div'
```
Could this be more future-proof ie maybe at some poitn we add `<button>`s ?

nerik (commented):
Good fix 👍

packages/react-hooks/src/use-map-interaction/use-map-interaction.ts
j8seangel: Good catch but I'd prefer at this stage (to production in 3 days) to cover only the use cases known

PR #712, opened 2021-07-14
MAP-456 Accept timebar locale prop

rdgfuentes:
- [x] Added optional `locale` prop to timebar to translate month names and day names in the timeline. Supported languages are: en, es, fr, id

### DIFF
--- a/packages/timebar/src/timebar.js
+++ b/packages/timebar/src/timebar.js
@@ -5,6 +5,10 @@ import dayjs from 'dayjs'
 import memoize from 'memoize-one'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import utc from 'dayjs/plugin/utc'
+import 'dayjs/locale/en'
+import 'dayjs/locale/es'
+import 'dayjs/locale/fr'
+import 'dayjs/locale/id'
 import ImmediateContext from './immediateContext'
 import {
   getTime,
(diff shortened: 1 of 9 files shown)


packages/timebar/src/timebar.js
```diff
@@ -86,9 +90,18 @@ class Timebar extends Component {
...
   })
 
-  componentDidMount() {
-    const { start, end } = this.props
+  componentDidUpdate(prevProps) {
+    const { locale: prevLocale } = prevProps
+    const { start, end, locale } = this.props
+    if (prevLocale !== locale) {
+      dayjs.locale(locale)
+      this.notifyChange(start, end, EVENT_SOURCE.TIME_RANGE_SELECTOR)
```
j8seangel, line 98: Are we sure we want to notify this change on locale update ?

packages/timebar/src/timebar.js
```diff
@@ -5,6 +5,10 @@ import dayjs from 'dayjs'
 import memoize from 'memoize-one'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import utc from 'dayjs/plugin/utc'
+import 'dayjs/locale/en'
```
j8seangel, line 8: would be nice import them dynamically on locale change ? so we could avoid some bundle size but I've reviewed and all of them are 12kb so just a nice to have if it is very easy change

j8seangel (commented):
Good work debugging and digging!

j8seangel:
Thanks for the update, much clearer now!

PR #732, opened 2021-07-21
use localStorage for categoryTab login link

j8seangel:
Save callbackUrl in local storage to avoid http 431 errors


satellitestudiodesign:
Works like a charm :) 

Even with crazy complex workspaces, like this with 7 filtered activity layers an 30 vessels
Screenshot 2021-07-21 at 13 24 43

PR #756, opened 2021-07-28
VV-123 Use default empty value for fields

rdgfuentes:


applications/vessel-history/src/features/profile/components/Info.tsx
```diff
@@ -159,20 +160,20 @@ const Info: React.FC<InfoProps> = (props): React.ReactElement => {
                 )}
               </div>
               <InfoField
-                vesselName={vessel.shipname ?? ''}
+                vesselName={vessel.shipname ?? DEFAULT_EMPTY_VALUE}
```
fertandil87, line 163: I see that vesselName is not used in InfoField, can be deleted?

applications/vessel-history/src/features/profile/components/Info.tsx
```diff
@@ -159,20 +160,20 @@ const Info: React.FC<InfoProps> = (props): React.ReactElement => {
...
                 label={VesselFieldLabel.builtYear}
-                value={vessel.builtYear ?? ''}
+                value={vessel.builtYear ?? DEFAULT_EMPTY_VALUE}
               ></InfoField>
               <InfoField
-                vesselName={vessel.shipname ?? ''}
+                vesselName={vessel.shipname ?? DEFAULT_EMPTY_VALUE}
                 label={VesselFieldLabel.owner}
-                value={vessel.owner ?? ''}
+                value={vessel.owner ?? DEFAULT_EMPTY_VALUE}
```
fertandil87, line 170: as you are doing {value.length > 0 ? value : DEFAULT_EMPTY_VALUE} in InfoField, all this default are not needed here

fertandil87 (commented):
@rdgfuentes

applications/vessel-history/src/features/profile/components/Info.tsx
rdgfuentes: No, because it's used to send to the InfoFieldHistory to set the title of the popup

applications/vessel-history/src/features/profile/components/Info.tsx
rdgfuentes: Good catch, I'll revert the change on those lines

applications/vessel-history/src/features/profile/components/Info.tsx
fertandil87: oh, I didn't see it, but I think is the same of the other comment so... we don't need to set every time we call the InfoField

PR #759, opened 2021-07-29
VV-126 fix fields info and historical

rdgfuentes:
- [x] Added AIS Transmission Dates at the end of the INFO Tab
- [x] Removed “Since [some date]”
- [x] Remove “+1 or +2 previous values” > Replace with “Historical values” that you can click and see the list of changes.
- [x] Display by default GFW data in info tab
- [x] When clicking in the “Historical values”, display all values: GFW + TMT data


applications/vessel-history/public/locales/source/translations.json
```diff
@@ -103,12 +103,14 @@
     "title": "Settings"
   },
   "vessel": {
+    "aisTransmissionPlural": "AIS Transmissions",
```
fertandil87, line 106: we are not using _plural?

fertandil87 (approved):

applications/vessel-history/public/locales/source/translations.json
rdgfuentes: good catch, will fix it

PR #761, opened 2021-07-30
added filters to the activity tab

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-131
image

- [x] Added modal with filters
- [x] reset filter every time a profile is opened
- [x] filter events based in the filters
- [x] add a new ButtonIcon design
- [x] add translations


applications/vessel-history/src/features/profile/filters/filters.slice.ts
```diff
@@ -0,0 +1,56 @@
...
+    },
+    resetFilters: (state, action: PayloadAction) => {
+      state.filters = initialState.filters
+    },
+  },
+})
+export default slice.reducer
+
+export const { updateFilters } = slice.actions
+export const { resetFilters } = slice.actions
```
rdgfuentes, line 46: You could combine these two exports into one 
```suggestion
export const { updateFilters, resetFilters } = slice.actions
```

applications/vessel-history/src/features/vessels/activity/vessels-activity.selectors.ts
```diff
@@ -0,0 +1,50 @@
+import { createSelector } from '@reduxjs/toolkit'
+import { RenderedEvent, selectEventsWithRenderingInfo } from 'features/vessels/activity/vessels-activity.slice';
+import { Filters, initialState, selectFilters } from './../../profile/filters/filters.slice';
```
rdgfuentes, line 3: It's preferred to use absolute path in our imports
```suggestion
import { Filters, initialState, selectFilters } from 'features/profile/filters/filters.slice';
```

applications/vessel-history/src/features/profile/components/activity/Activity.tsx
```diff
@@ -1,13 +1,12 @@
...
 import { selectResourcesLoading } from 'features/resources/resources.slice'
 import { VesselWithHistory } from 'types'
-import {
-  RenderedEvent,
-  selectEventsWithRenderingInfo,
-} from 'features/vessels/activity/vessels-activity.slice'
+import { RenderedEvent} from 'features/vessels/activity/vessels-activity.slice'
+import { selectFilteredEvents } from 'features/vessels/activity/vessels-activity.selectors'
 import { fetchRegionsThunk } from 'features/regions/regions.slice'
+import ActivityFilters from '../../filters/ActivityFilters'
```
rdgfuentes, line 9:
```suggestion
import ActivityFilters from 'features/profile/filters/ActivityFilters'
```

applications/vessel-history/src/features/vessels/activity/vessels-activity.selectors.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const selectFilteredEvents = createSelector(
+  [selectEventsWithRenderingInfo, selectFilters],
+  (events, filters) => {
+    const startTimestamp = new Date(filters.start).getTime()
+    const endTimestamp = new Date(filters.end).getTime()
```
rdgfuentes, line 9: When defining a date range to filter on, remember to add the proper time on the start and end. In this case:

- `startTimestamp` should be calculated for the start date selected by the user and add the `00:00:00.000` as the time part
- `endTimestamp` should be calculated for the end date selected by the user and add the `23:59:59.999` as the time part
So that any event that took place in the startDate or endDate selected by the user is shown

applications/vessel-history/src/features/vessels/activity/vessels-activity.selectors.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    return events.flat().filter((event: RenderedEvent) => {
+      if (startTimestamp > event.start || endTimestamp < event.end) {
```
rdgfuentes, line 12: Include the event if took place in the same timestamp of start or end too
```suggestion
      if (startTimestamp >= event.start || endTimestamp <= event.end) {
```

applications/vessel-history/src/features/profile/filters/filters.hooks.ts
```diff
@@ -0,0 +1,43 @@
+import { useDispatch, useSelector } from 'react-redux'
+import {
+  selectFilters,
+  updateFilters,
+  availableEventFilters,
+} from './filters.slice'
+
+export const useApplyFiltersConnect = () => {
```
rdgfuentes, line 8: This is so clean! I really like it! 👍

applications/vessel-history/src/features/profile/filters/ActivityFilters.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <br/>
+          <InputDate 
+            value={start} 
+            onChange={(e) => {
+              if (e.target.value !== start) {
+                setDate('start', e.target.value )
+              }
+            }}
+            onRemove={() => {
+              setDate('start', DEFAULT_WORKSPACE.start)
```
rdgfuentes, line 50: I think the default start here should be the date of the first event of the vessel

applications/vessel-history/src/features/profile/filters/ActivityFilters.tsx
```diff
@@ -0,0 +1,79 @@
...
+            max={DEFAULT_WORKSPACE.end} />
+          <InputDate 
+            value={end} 
+            onChange={(e) => {
+              if (e.target.value !== end) {
+                setDate('end', e.target.value )
+              }
+            }}
+            onRemove={() => {
+              setDate('end', DEFAULT_WORKSPACE.end)
```
rdgfuentes, line 62: Maybe the default end here should be the date of the last event of activity of the vessel. Follow up the comment I added on the story asking Gise about this.

rdgfuentes (commented):
I really like how you structured the components and hooks in this approach 👏 . Despite those minor comments I added, I noticed the performance is not the best, the UI of the filters  freezes when [opening a vessel with lots of activity events](https://localhost:3000/profile/public-global-carrier-vessels%3Av20201001/fbd0ce041-1286-bbcc-e9d8-0e0247265d9d/NA?latitude=65.999103&longitude=12.264913&zoom=8).
In regards to the performance issue something that could help is inverting the order in which we are using the selectors to build the list of events to show. Today we are 
```
1) Selecting All Events 
2) Attaching additional event information to them
3) Filtering out events based on the filters
```
 I'd suggest to change the order and do it this way instead: 
```
1) Selecting All Events 
2) Filtering out events based on the filters
3) Attaching additional event information to them
```
so that we are only attaching info to the events we need to display.

applications/vessel-history/src/features/profile/filters/ActivityFilters.tsx
fertandil87: yes, but the workspace start / end should be the same of the start and end that we use to get the events

PR #773, opened 2021-08-06
VV-137 select vessel events onclick

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-137
- [x] Select vessel events when clicking on them in the map tab of the vessel profile
- [x] Customized vessel events styles on the map: Not switching the events points to segments, custom stroke color and icon size depending on its status (active/inactive)
image

### DIFF
--- a/packages/layer-composer/src/generators/vessel-events/vessel-events.ts
+++ b/packages/layer-composer/src/generators/vessel-events/vessel-events.ts
@@ -33,6 +33,7 @@ class VesselsEventsGenerator {
     return (
       config.track &&
       config.pointsToSegmentsSwitchLevel !== undefined &&
+      Number.isFinite(config.pointsToSegmentsSwitchLevel) &&
       (config.zoom as number) >= config.pointsToSegmentsSwitchLevel
     )
   }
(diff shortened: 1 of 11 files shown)


packages/layer-composer/src/generators/vessel-events/vessel-events.ts
```diff
@@ -24,14 +24,18 @@ interface VesselsEventsSource extends GeoJSONSourceRaw {
 
 export type GlobalVesselEventsGeneratorConfig = MergedGeneratorConfig<VesselEventsGeneratorConfig>
 
-const POINTS_TO_SEGMENTS_ZOOM_LEVEL_SWITCH = 9
```
fertandil87, line 27: and where is this default value to keep working the other projects?

rdgfuentes: With @j8seangel we decided to change the default behavior to not switch to segments view here in the `vessel-events.ts` but he already modified the dataview of the vessel events to have this value set to 9 in both `dev` and `pro` so the current behavior is not changed in `fishing-map`:
```
        "config": {
            "type": "TRACK",
            "color": "#F95E5E",
            "showIcons": false,
            "showAuthorizationStatus": false,
            "pointsToSegmentsSwitchLevel": 9
        },
```

fertandil87 (approved):

PR #775, opened 2021-08-06
fix events in tmt vessels

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-150

Fix error in TMT vessels after join in a GFW vessel (wrong activities)


rdgfuentes (approved):
👍

PR #789, opened 2021-08-10
added iuu status

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-132

image

### DIFF
--- a/applications/vessel-history/src/features/profile/components/Info.tsx
+++ b/applications/vessel-history/src/features/profile/components/Info.tsx
@@ -222,6 +222,14 @@ const Info: React.FC<InfoProps> = (props): React.ReactElement => {
                   
                 </div>
               </div>
+              {vessel.iuuStatus !== undefined && 
+                <InfoField
+                  vesselName={vessel.shipname ?? DEFAULT_EMPTY_VALUE}
+                  label={VesselFieldLabel.iuuStatus}
+                  value={t(`vessel.iuuStatusOptions.${vessel.iuuStatus}` as any, t('common.unknown', 'unknown'))}
+                  valuesHistory={[]}
+                ></InfoField>
+              }
             </div>
           </Fragment>
         )}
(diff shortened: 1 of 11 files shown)


applications/vessel-history/src/features/profile/components/Info.tsx
```diff
@@ -201,6 +201,14 @@ const Info: React.FC<InfoProps> = (props): React.ReactElement => {
                   )}
                 </div>
               </div>
+              {vessel.iuuStatus !== undefined &&
```
rdgfuentes, line 204: Here the IUU Status will not be shown if we don't have TMT data, that's not correct. Because if we don’t have TMT data, IUU listed status should display `unknown`

applications/vessel-history/src/features/profile/components/Info.tsx
```diff
@@ -201,6 +201,14 @@ const Info: React.FC<InfoProps> = (props): React.ReactElement => {
                   )}
                 </div>
               </div>
+              {vessel.iuuStatus !== undefined && 
+                <InfoField
+                  vesselName={vessel.shipname ?? DEFAULT_EMPTY_VALUE}
+                  label={VesselFieldLabel.iuuStatus}
+                  value={t(`vessel.iuuStatusOptions.${vessel.iuuStatus}` as any, t('common.unknown', 'unknown'))}
```
rdgfuentes, line 208: In this case if we don't have the translation for that IUU status we could probably display the IUU status code , [follow up on the comment I added to the ticket](https://globalfishingwatch.atlassian.net/browse/VV-132?focusedCommentId=13395) about this.

rdgfuentes (changes requested):
Some minor changes to display always the field in the info tab and the default text to display when the iuu status code is not found in the translations, the rest looks good 👍

applications/vessel-history/src/features/profile/components/Info.tsx
fertandil87: I fixed both comment in one place

PR #796, opened 2021-08-13
VV-68 Display port visits in activity and map

rdgfuentes:
- Created Vessels dataview to use in Vessel Viewer
  -  Reviewed and fixed all `datasetsConfig` in it
- Added new `port-visit` event type (left `port` in case that's been used by other apps like CVP)
- Added new icon for port visit events
- Styled port visit events appropriately 
- Generated event description based on anchorage name and location when available.
- Display port visit event info popup
- Added support for PSMA (fetching the list from an API endpoint to be implemented when the endpoint is ready)
- Updated tests

image
image
image


packages/api-types/src/events.ts
```diff
@@ -113,6 +134,7 @@ export interface ApiEvent<Vessel = EventVessel> {
   loitering?: LoiteringEvent
   encounter?: EncounterEvent<Vessel>
   port?: PortEvent
+  port_visit?: PortVisitEvent
```
j8seangel, line 137: we using snake_case here but camelCase in the rest of the properties, is this a typo ?

j8seangel (approved):
LGTM, just a question about the api port type field

packages/api-types/src/events.ts
rdgfuentes: No, it is not a typo. This must match the event property returned by the API otherwise the data for port visit found in the event.

PR #803, opened 2021-08-16
[WIP] resources query refacto

nerik:
The main idea is to remove app logic in `dataviews-client` when preparing resource queries from url dataviews.
It is done by:
- app side: manipulate datasetConfigs:
   - collect only relevant info, track and event datasetConfigs
   - filter out event datasetConfigs without vessel query param
   - [TODO] modify track thinning query params
   - [TODO] add extra datasetConfig for loading tyrack speed
- dataviews-client side : prepare resource queries (ie resolve urls) based on all dataviews' available datasetConfigs


applications/fishing-map/src/features/resources/resources.selectors.ts
```diff
@@ -20,7 +26,51 @@ export const selectVisibleResources = createSelector(
...
 
-export const selectDataviewsResourceQueries = createSelector(
+/**
+ * Prepare dataviews for querying resources, by altering dataset configs
+ * - Filter out non track dvs
+ * - Add thinning query params
+ * - Add extra dataset config for track speed, if needed
+ * - Filter events dataset configs
+ */
+export const selectDataviewsForResourceQuerying = createSelector(
```
rdgfuentes, line 36: Would be possible to extract the logic of this selector out to the dataviews-client package so that it can be reused also in other apps consuming dataviews like vessel-viewer?

nerik: It would be good to discuss this. Right now I'm bothered by the fact that dataviews-client handles what in my view is application logic:
- which type of resources do we want to load from a track dataview?
- what thinning parameters do we want to apply
- what fields do we want to load on the track?
- etc

What can be done though is move some of the barebones dataviews datasetsConfig modification logic as an helper in dataviews-client (in the spirit of https://github.com/GlobalFishingWatch/frontend/pull/803/files#diff-8a29c0005ba549ec031b20d1e6579e73cd0cd5c9e39406770f548b1d9980a442R71)

nerik: > What can be done though is move some of the barebones dataviews datasetsConfig modification logic as an helper in dataviews-client

Something like `getBaseTrackDataviewDatasetConfigs` if yo know what I mean

rdgfuentes: Makes sense moving them to a helper in dataviews-client that could accept those per app configurations as arguments (type of resources, thinning parameters, etc..)

packages/dataviews-client/src/resolve-dataviews.ts
```diff
@@ -68,39 +68,141 @@ export const mergeWorkspaceUrlDataviewInstances = (
...
+  return {
+    info,
+    track,
+    events,
+  }
+}
+
+export const getDataviewsForResourceQuerying = (
+  dataviewInstances: UrlDataviewInstance[],
+  trackDatasetConfigsTransform?: (
```
j8seangel, line 122: This seems to be very adhoc for the track datasets... and the idea is to keep it as generic as possible. so why not having something like `DatasetConfigsTransform` being a dictionary of dataview.type as keys and the function as the value, this way we could use the function in any dataview type

packages/dataviews-client/src/resolve-dataviews.ts
```diff
@@ -68,39 +68,141 @@ export const mergeWorkspaceUrlDataviewInstances = (
...
+) => {
+  if (!dataviewInstances) return []
+  const preparedDataviewsInstances = dataviewInstances.map((dataviewInstance) => {
+    if (dataviewInstance.config?.type !== Generators.Type.Track) return dataviewInstance
+
+    const { track, info, events } = getTrackDataviewDatasetConfigs(dataviewInstance)
+    let preparedDatasetConfigs
+
+    if (trackDatasetConfigsTransform) {
+      preparedDatasetConfigs = trackDatasetConfigsTransform({ track, info, events })
```
j8seangel, line 134: if we go to the generic callback this doesn't scale so I'd return here the same format than the original response, that said I'd return an array instead of an object

packages/dataviews-client/src/resolve-dataviews.ts
```diff
@@ -68,39 +68,141 @@ export const mergeWorkspaceUrlDataviewInstances = (
...
-
-  const url = resolveEndpoint(dataset, datasetConfig)
+  const resources = datasets?.flatMap((dataset) => {
+    const datasetConfig = dataview?.datasetsConfig?.find(
+      (datasetConfig) => datasetConfig.datasetId === dataset.id
+    )
+    if (!datasetConfig) return []
+    const url = resolveEndpoint(dataset, datasetConfig)
+    if (!url) return []
+    return [{ dataset, datasetConfig, url, dataviewId: dataview.dataviewId as number }]
```
j8seangel, line 192: I like having the dataviewId associated in the resource, this will simplify A LOT reading resources.
Thanks!

applications/fishing-map/src/features/workspace/Workspace.tsx
```diff
@@ -117,7 +118,9 @@ function Workspace() {
   const workspace = useSelector(selectWorkspace)
   const workspaceStatus = useSelector(selectWorkspaceStatus)
   const locationCategory = useSelector(selectLocationCategory)
-  const resourceQueries = useSelector(selectDataviewsResourceQueries)
+  const resourceQueries = resolveResourcesFromDatasetConfigs(
+    useSelector(selectDataviewsForResourceQuerying) ?? []
+  )
 
   useEffect(() => {
```
j8seangel, line 125: As we are refactoring this... what about having a validation to avoid requesting the same request already done with an error every time another resource changes? 
What happens now is:
1. You select a vessel and the track fails
2. You select another vessel
3. The first vessel tries to reload the track again

applications/fishing-map/src/features/workspace/Workspace.tsx
```diff
@@ -117,7 +118,9 @@ function Workspace() {
   const workspace = useSelector(selectWorkspace)
   const workspaceStatus = useSelector(selectWorkspaceStatus)
   const locationCategory = useSelector(selectLocationCategory)
-  const resourceQueries = useSelector(selectDataviewsResourceQueries)
+  const resourceQueries = resolveResourcesFromDatasetConfigs(
```
j8seangel, line 121: This is going to end up having a new array every time the component is rendered, so why not having a selector which return this function?

applications/fishing-map/src/features/resources/resources.selectors.ts
```diff
@@ -20,7 +21,13 @@ export const selectVisibleResources = createSelector(
   }
 )
 
-export const selectDataviewsResourceQueries = createSelector(
-  [selectDataviewInstancesResolved],
-  (dataviewInstances) => resolveDataviewsResourceQueries(dataviewInstances ?? [])
+export const selectThinningConfig = createSelector(
```
j8seangel, line 24: Much simpler, thanks!

packages/dataviews-client/src/resources/resources-resolvers.ts
```diff
@@ -6,6 +6,7 @@ import {
   UrlDataviewInstance,
 } from '../resolve-dataviews'
 
+// TODO Deprecate
```
j8seangel, line 9: Aren't we removing this because it is already in use for vessel-examiner?

j8seangel (commented):
Would be nice to talk about how to make the transform callback generic, the rest looks nice.

Also seeing the dataviewId into the resource made me realised maybe it is easier to grab the resource in components like [VesselLayerPanel](https://github.com/GlobalFishingWatch/frontend/blob/resources-query-refacto/applications/fishing-map/src/features/workspace/vessels/VesselLayerPanel.tsx#L39) without the need of the `resolveDataviewDatasetResource` WDYT ?

packages/dataviews-client/src/resolve-dataviews.ts
```diff
@@ -110,32 +104,32 @@ const getTrackDataviewDatasetConfigs = (
...
-    if (dataviewInstance.config?.type !== Generators.Type.Track) return dataviewInstance
-
-    const { track, info, events } = getTrackDataviewDatasetConfigs(dataviewInstance)
     let preparedDatasetConfigs
+    switch (dataviewInstance.config?.type) {
+      case Generators.Type.Track:
+        preparedDatasetConfigs = getTrackDataviewDatasetConfigs(dataviewInstance)
+        if (datasetConfigsTransform && datasetConfigsTransform[Generators.Type.Track]) {
+          preparedDatasetConfigs =
+            datasetConfigsTransform[Generators.Type.Track]!(preparedDatasetConfigs)
```
j8seangel, line 126: Isn't Eslint complaining about using [Forbidden non-null assertion](https://github.com/typescript-eslint/typescript-eslint/blob/v4.29.1/packages/eslint-plugin/docs/rules/no-non-null-assertion.md) ?
why not using `?.` instead ? what is the difference ?

packages/dataviews-client/src/resolve-dataviews.ts
```diff
@@ -110,32 +104,32 @@ const getTrackDataviewDatasetConfigs = (
...
 
-  return {
-    info,
-    track,
-    events,
-  }
+  return [info, track, ...events]
 }
 
+export type DatasetConfigsTransforms = Partial<
```
j8seangel, line 110: Thanks! this should be easier to use in our apps for other datasets types

j8seangel (commented):
Just a small comment, thanks for the changes!

packages/dataviews-client/src/resources/resources-resolvers.ts
nerik: See https://github.com/GlobalFishingWatch/frontend/pull/820

PR #815, opened 2021-08-18
Fishing map/private datasets

j8seangel:
- Render lock icon in private datasets:
image
- New section in user panel to render the private workspaces checking user group and private workspaces with that id (only visible when there are workspaces)
image


enriquetuya:
@j8seangel @satellitestudiodesign I would add the lock on the left side, not on the right. I think having it on the left makes it easier for users to see which ones are private with a quick look if they have several private layers.

applications/fishing-map/src/features/datasets/datasets.utils.ts
```diff
@@ -18,6 +19,13 @@ export const removeDatasetVersion = (datasetId: string) => {
   return datasetId ? datasetId?.split(':')[0] : ''
 }
 
+export const getDatasetLabel = (dataset: { id: string; name?: string }): string => {
+  const { id, name = '' } = dataset || {}
+  if (!id) return name || ''
+  const label = getDatasetNameTranslated(dataset)
+  return dataset.id.includes(PRIVATE_SUFIX) ? `${label} 🔒` : label
```
nerik, line 26: Why displaying an emoji when we have a good-looking icon? 
--> https://github.com/GlobalFishingWatch/frontend/pull/815/files#diff-e1152028eb6a3aeb8f5cd07e285f5f727f76aff26dd2dfe1f9f8d48eb15b124bR60

nerik: This method could optionally return a React element when it's possible to use one? 
It would be possible, for instance, here: https://github.com/GlobalFishingWatch/frontend/pull/815/files#diff-3145ce8092c05120599c3ac73fd565cb66ad5ca0f456527adeed3988ad638bb1R233

When it's not, I'd much rather have `${label} (private)`, than the emoji

applications/fishing-map/src/features/user/UserWorkspacesPrivate.tsx
```diff
@@ -0,0 +1,72 @@
...
+                    category: workspace.category || WorkspaceCategories.FishingActivity,
+                    workspaceId: workspace.id,
+                  },
+                  query: {},
+                }}
+                onClick={() => onWorkspaceClick(workspace)}
+              >
+                <Icon icon="private" />
+                <span className={styles.workspaceTitle}>{workspace.name}</span>
+                <IconButton icon="arrow-right" />
```
nerik, line 62: Is there any particular reason why delete and rename buttons are not available for private workspaces?

applications/fishing-map/src/features/dataviews/dataviews.utils.ts
```diff
@@ -140,3 +142,8 @@ export const getContextDataviewInstance = (
   }
   return contextDataviewInstance
 }
+
+export const dataviewWithPrivateDatasets = (dataview: UrlDataviewInstance) => {
+  const datasets = dataview.datasets || []
+  return datasets.some((d) => d.id.includes(PRIVATE_SUFIX))
```
nerik, line 148: Sorry if already asked/obvious: why can't we use a `private` flag on datasets and workspaces?

If not possible, could we at least try to centralize the logic here with https://github.com/GlobalFishingWatch/frontend/pull/815/files#diff-7f372e623785bb5176dfba7113a764c9456a03c8c3707f60c87567f8d5b5258dR26, and also in `user.selectors`

nerik (changes requested):
Please fix the issue with the dataviews-client path, the rest is less important.
Otherwise, that's a great addition, thanks 👏

applications/fishing-map/src/features/dataviews/dataviews.utils.ts
```diff
@@ -7,13 +7,15 @@ import {
...
 import { Generators } from '@globalfishingwatch/layer-composer'
+import { PRIVATE_SUFIX } from 'data/config'
 import {
   TEMPLATE_ENVIRONMENT_DATAVIEW_ID,
   TEMPLATE_CONTEXT_DATAVIEW_ID,
   DEFAULT_FISHING_DATAVIEW_ID,
   TEMPLATE_VESSEL_DATAVIEW_ID,
   TEMPLATE_USER_TRACK_ID,
 } from 'data/workspaces'
+import { UrlDataviewInstance } from '../../../../../packages/dataviews-client/dist'
```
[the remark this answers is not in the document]
j8seangel: 🤦 sorry! I don't know why it compiled and didn't show any error but already fixed! thanks

applications/fishing-map/src/features/datasets/datasets.utils.ts
j8seangel: I tried first to use the <Icon /> component but there are many places where the component API supports only a string, for example:
- [`<TagList />`](https://github.com/GlobalFishingWatch/frontend/blob/fishing-map/private-datasets/packages/ui-components/src/tag-list/index.ts#L8) used in sources filter layer panel:
image

- [`<MultiSelect />`](https://github.com/GlobalFishingWatch/frontend/blob/fishing-map/private-datasets/packages/ui-components/src/multi-select/index.ts#L6) used in the search:
image

So we thought this was the easier way to include it as we can keep as plain text. 
But we could support ReactElements if we think this can be useful in the future, WDYT ?

applications/fishing-map/src/features/user/UserWorkspacesPrivate.tsx
j8seangel: Because you might not be the creator of it. Only the owner can do it and if that is the case it will appear in the list below

applications/fishing-map/src/features/dataviews/dataviews.utils.ts
j8seangel: Merged in https://github.com/GlobalFishingWatch/frontend/pull/815/commits/9fb6972a546bdbc5470528b475441c1f5aba9bab

applications/fishing-map/src/features/datasets/datasets.utils.ts
nerik: I think that would be worth it and not too hard to do (only a type change in the best case scenario)

j8seangel: Sounds reasonable, let me try it and will let you know!

PR #816, opened 2021-08-18
VV-153 Highlight configuration - update some settings

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-153


applications/vessel-history/public/locales/source/translations.json
```diff
@@ -10,7 +10,10 @@
     "since": "since",
     "timeRange": "time range",
     "to": "to",
+    "typeToSeach": "Type to search",
```
fertandil87, line 13: delete this

applications/vessel-history/src/features/settings/settings.hooks.ts
```diff
@@ -46,3 +53,82 @@ export const useApplySettingsConnect = () => {
...
+    [section, setSettingOptions, anyOption]
+  )
+
+  const getOptions = useCallback(
+    (availableOptions: MultiSelectOption[], field: string, selected?: string | string[]) => {
+      const allOptions = [anyOption, ...availableOptions]
+      const selectedOptions = allOptions.filter((option) => selected?.includes(option.id))
+      const options = [
+        // First display ANY whether it's selected or not
+        anyOption,
```
fertandil87, line 104: what happen if any is selected, and any other random option? is like all are selected so if any is selected we can avoid display the list, can that help to the performance? is an idea

PR #834, opened 2021-08-23
Adds a yearly timeline of transmissions to vessels search card

satellitestudiodesign:
Screenshot 2021-08-23 at 18 05 56

### DIFF
--- a/applications/fishing-map/src/data/config.ts
+++ b/applications/fishing-map/src/data/config.ts
@@ -45,14 +45,17 @@ export const DEFAULT_TIME_RANGE = {
 
 export const DEFAULT_ACTIVITY_CATEGORY = 'fishing'
 
+export const FIRST_YEAR_OF_DATA = 2012
+export const CURRENT_YEAR = new Date().getFullYear()
+
 export const DEFAULT_WORKSPACE = {
   ...DEFAULT_VIEWPORT,
   query: undefined,
   readOnly: false,
   daysFromLatest: undefined,
   sidebarOpen: true,
-  availableStart: new Date(Date.UTC(2012, 0, 1)).toISOString(),
-  availableEnd: new Date(Date.UTC(new Date().getFullYear(), 11, 31)).toISOString(),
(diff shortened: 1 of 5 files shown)


applications/fishing-map/src/data/config.ts
```diff
@@ -45,14 +45,17 @@ export const DEFAULT_TIME_RANGE = {
 
 export const DEFAULT_ACTIVITY_CATEGORY = 'fishing'
 
+export const FIRST_YEAR_OF_DATA = 2012
+export const CURRENT_YEAR = new Date().getFullYear()
```
nerik, line 49: Thanks 👍

nerik (commented):
Great addition 👏 
Two comments:
- I have proposed a simplification of the year counter code: https://github.com/GlobalFishingWatch/frontend/pull/834/commits/4e666c6e085722db25e2b36e43cc4095314bb603
- I have an issue with the design where I didn't understand what the numbers '16', '17', etc meant, until I looked at the code 😞
Probably lack of context and/or sleep on my side, but here's an option that could work IMHO:
Screenshot 2021-08-24 at 11 43 05
Maybe the actual mini timebar is overkill, but the idea is to strike a balance between showing full years (so that idiots can understand what the numbers mean) and not having all the years with full numbers which is noisy.

applications/fishing-map/src/features/search/Search.tsx
```diff
@@ -420,12 +421,16 @@ function Search() {
...
                               )}
                               {firstTransmissionDate && lastTransmissionDate && (
-                                <div className={styles.property}>
+                                <div className={cx(styles.property, styles.fullWidth)}>
                                   <label>{t('vessel.transmission_plural', 'Transmissions')}</label>
                                   <span>
                                     from <I18nDate date={firstTransmissionDate} /> to{' '}
                                     <I18nDate date={lastTransmissionDate} />
                                   </span>
+                                  <TransmissionsTimeline
```
j8seangel, line 430: There are some datasets with no transmission data... so I'd not render the component if the data is not available to avoid the empty timeline:
image

PR #870, opened 2021-09-01
Fishing map/polish

j8seangel:
- [x] Show dataset as source in vessel info to match datasets and translate it
image

- [x] Centralice activity dataviews naming and include missing translations
Before: 
image
Now:
image

- [x] Improve sources selection filter. Keeping the current selection when all are selected instead of marking only the "allOption" avoids a weird jump.


nerik (approved):
Excellent, LGTM 👍

PR #878, opened 2021-09-03
User access

nerik:
See https://globalfishingwatch.atlassian.net/browse/VV-146

Screenshot 2021-09-03 at 12 12 15


rdgfuentes (approved):
LGTM  🚀

PR #879, opened 2021-09-03
VV130 Group vessel events by voyages

rdgfuentes:
- [x] Group vessel events by voyages based on port visits
- [x] Voyages are still displayed even if port visits are hidden in the filters
- [x] Voyages will only appear if there are events to show in them based in the filters
- [x] Split voyages into entered and exit port events
image


applications/vessel-history/src/features/profile/components/activity/ActivityVoyage.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, [event.from, event.start, event.to, event.end, t])
+
+  return (
+    <Fragment>
+      <div className={cx(styles.event, styles.voyage)}>
+        <div className={styles.eventData}>
+          <div className={styles.description}>{voyageLabel}</div>
+        </div>
+        <div className={styles.actions}>
+          <IconButton
```
fertandil87, line 67: hide buttons if there is not events inside because they are filtered

fertandil87 (changes requested):

applications/vessel-history/src/features/profile/components/activity/ActivityVoyage.tsx
rdgfuentes: good catch! fixed

PR #896, opened 2021-09-08
Vessel history/perf improvements

j8seangel:
Improve VV performance with:
- [x] Not rendering all tabs until is loaded, for example this avoid rendering the map (expensive in memory and network) until needed
- [x] Avoid unnecessary and heavy re-renders


j8seangel:
Closing in favour of https://github.com/GlobalFishingWatch/frontend/pull/897 cherry-picking perf commits

PR #897, opened 2021-09-08
Vessel history/performance improvements

j8seangel:
Improve VV performance with:
- [x] Not rendering all tabs until is loaded, for example this avoid rendering the map (expensive in memory and network) until needed
- [x] Avoid unnecessary and heavy re-renders


applications/vessel-history/src/features/settings/settings.hooks.ts
```diff
@@ -54,20 +54,17 @@ export const useSettingsConnect = () => {
   }
 }
 
-export const useSettingsRegionsConnect = (
-  section: SettingEventSectionName,
-  settings: SettingsEvents
-) => {
+const onlyUnique = (value: Region, index: number, self: Region[]) => {
```
rdgfuentes, line 57: 👍  Right, we don't need to create a new instance of this on every call to the `useSettingsRegionsConnect` hook

applications/vessel-history/src/features/regions/regions.slice.ts
```diff
@@ -60,9 +62,9 @@ export const fetchRegionsThunk = createAsyncThunk(
         options
       )
       const result: Regions[] = [
-        { id: MarineRegionType.eez, data: eezs },
-        { id: MarineRegionType.mpa, data: mpas },
-        { id: MarineRegionType.rfmo, data: rfmos },
+        { id: MarineRegionType.eez, data: eezs.sort(sortRegionAlphabetically) },
```
rdgfuentes, line 65: 💯

PR #927, opened 2021-09-15
make public workspaces optional

j8seangel:
Replace the window.alert in workspace save for the following modal to add the public option:

image

and disable it when private datasets are in use:
image


applications/fishing-map/src/features/sidebar/SidebarHeader.tsx
```diff
@@ -21,94 +17,70 @@ import { AsyncReducerStatus } from 'utils/async-slice'
...
   const workspaceCustomStatus = useSelector(selectWorkspaceCustomStatus)
   const { showClipboardNotification, copyToClipboard } = useClipboardNotification()
   const workspace = useSelector(selectWorkspace)
 
   const isOwnerWorkspace = workspace?.ownerId === userData?.id
-  const isDefaultWorkspace = workspace?.id === DEFAULT_WORKSPACE_ID
 
-  const updateWorkspaceAction = (workspaceId: string) => {
-    return dispatch(updatedCurrentWorkspaceThunk(workspaceId))
+  const onCloseCreateWorkspace = () => {
```
nerik, line 44: Shouldn't we use `useCallback` ?

applications/fishing-map/src/features/workspace/shared/NewWorkspaceModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+      const start = formatI18nDate(timerange.start as string, {
+        format: dateFormat,
+      })
+        .replace(',', '')
+        .replace('.', '')
+      const end = formatI18nDate(timerange.end as string, {
+        format: dateFormat,
+      })
+        .replace(',', '')
+        .replace('.', '')
```
nerik, line 50:
```suggestion
      const formatTimerangeBoundary = (boundary?: string) => {
        return formatI18nDate(boundary as string, {
        format: dateFormat,
      })
        .replace(/[\.\,]/g, '')
      }
      const start = formatTimerangeBoundary(timerange.start)
      const end = formatTimerangeBoundary(timerange.end)
```

applications/fishing-map/src/features/workspace/shared/NewWorkspaceModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+      onClose={onClose}
+    >
+      <InputText
+        inputSize="small"
+        value={name}
+        label={t('common.name', 'Name')}
+        className={styles.input}
+        onChange={(e) => setName(e.target.value)}
+      />
+      <Radio
```
nerik, line 98: This must be a `Checkbox`, not a `Radio`

nerik (changes requested):
Looks good! Please see my comments.

I have doubts about the flow: when a user saves a workspace once, the following saves won't trigger the modal again. Which means there is no way to change the public visibility setting again, nor changing the title (which means you can completely change dates for instance and still have the old dates in the title). Is that intended? How do we allow the user to create a new workspace from an existing one, rather than updating it?

applications/fishing-map/src/features/workspace/shared/NewWorkspaceModal.tsx
j8seangel: Thanks, added in https://github.com/GlobalFishingWatch/frontend/commit/264d9caefaa5e320a92028293a3f2e1b6d1adb75

applications/fishing-map/src/features/workspace/shared/NewWorkspaceModal.tsx
j8seangel: Replaced by a Switch, we don't have a Checkbox component yet

PR #962, opened 2021-09-21
VV-205 Support TMT special dates with only year or only year and month

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-205
- [x] Added support for special TMT dates to display year or year and month when it's not exact
- [x] Ensure authorisation list is sorted and formatted properly when dealing with TMT special dates

image


applications/vessel-history/src/features/profile/components/InfoField.tsx
```diff
@@ -35,31 +35,27 @@ interface ListItemProps {
   value?: string
   valuesHistory?: ValueItem[]
   vesselName: string
-  hideTMTDate?: boolean
```
fertandil87, line 38: why are you deleting all hideTMTDate?

fertandil87 (changes requested):

applications/vessel-history/src/features/profile/components/InfoField.tsx
rdgfuentes: Because this story is to display all TMT dates based on the special cases like 19970000 and 19970900

rdgfuentes: Good catch @fertandil87 ! You are right, we should continue hiding the dates on those specific fields. Thanks for catching it!

PR #972, opened 2021-09-22
VV-246 Display all event regions

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-246
- [x] Display all regions where the event happened

image


satellitestudiodesign (approved):
👏  Looks great!

PR #973, opened 2021-09-22
VV-239 Refresh token automatically

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-239
- [x] Ensure the user has a valid token before performing the vessel search
- [x] Stop using useGFWLogin in multiple components to reduce the number of successive unnecessary calls to `/auth/me`
- [x] Enhanced user hook to share its state thru redux centralizing user data and actions
- [x] Refactored advanced search logic into a hook
- [x] Fixed warning in AdvancedSearch `A component is changing an uncontrolled input to be controlled.` 
- [x] Fixed analytics to not track user refresh data & token as login event

### DIFF
--- a/applications/vessel-history/src/features/home/Home.tsx
+++ b/applications/vessel-history/src/features/home/Home.tsx
@@ -1,10 +1,11 @@
-import React, { Fragment, useCallback, useEffect, useRef } from 'react'
+import React, { Fragment, useCallback, useEffect } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { useTranslation } from 'react-i18next'
 import { event as uaEvent } from 'react-ga'
 import Link from 'redux-first-router-link'
 import { redirect } from 'redux-first-router'
 import { DateTime, Interval } from 'luxon'
+import GFWAPI from '@globalfishingwatch/api-client'
 import { VesselSearch } from '@globalfishingwatch/api-types'
 import Logo from '@globalfishingwatch/ui-components/dist/logo'
 import { Spinner, IconButton, Button } from '@globalfishingwatch/ui-components'
@@ -13,12 +14,7 @@ import VesselListItem from 'features/vessel-list-item/VesselListItem'
 import { useOfflineVesselsAPI } from 'features/vessels/offline-vessels.hook'
 import { selectAll as selectAllOfflineVessels } from 'features/vessels/offline-vessels.slice'
 import SearchPlaceholder, { SearchNoResultsState } from 'features/search/SearchPlaceholders'
-import {
-  selectAdvancedSearchFields,
-  selectHasSearch,
-  selectUrlQuery,
-} from 'routes/routes.selectors'
-import { fetchVesselSearchThunk } from 'features/search/search.thunk'
+import { selectHasSearch } from 'routes/routes.selectors'
 import {
   selectSearchOffset,
   selectSearchResults,
@@ -28,7 +24,7 @@ import {
 import AdvancedSearch from 'features/search/AdvancedSearch'
 import { useUser } from 'features/user/user.hooks'
 import { PROFILE } from 'routes/routes'
-import { useSearchConnect } from 'features/search/search.hooks'
+import { useSearchConnect, useSearchResultsConnect } from 'features/search/search.hooks'
 import { formatVesselProfileId } from 'features/vessels/vessels.utils'
 import styles from './Home.module.css'
 import LanguageToggle from './LanguageToggle'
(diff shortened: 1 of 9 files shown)


applications/vessel-history/src/features/home/Home.tsx
```diff
@@ -101,28 +93,6 @@ const Home: React.FC<LoaderProps> = (): React.ReactElement => {
     if (selectedVessel) openVesselProfile(selectedVessel, akaVessels)
   }, [openVesselProfile, selectedVessels, vessels])
 
-  const fetchResults = useCallback(
```
j8seangel, line 104: way cleaner! 👍

applications/vessel-history/src/features/search/AdvancedSearch.tsx
```diff
@@ -60,6 +47,13 @@ const AdvancedSearch: React.FC = () => {
     [dispatchQueryParams]
   )
 
+  const onSearchClick = useCallback(
```
j8seangel, line 50: why do we need to create a memoized function for something which doesn't have params?
I'd use `fetchResults` directly [here](https://github.com/GlobalFishingWatch/frontend/pull/973/files#diff-3f1966ff8a700843a0d4e57620497c7b9f85614237dbb24c3a1c0f74981747afR161) to simplify it

applications/vessel-history/src/features/search/AdvancedSearch.tsx
```diff
@@ -80,7 +74,7 @@ const AdvancedSearch: React.FC = () => {
         <InputText
           inputSize="small"
           onChange={onMainQueryChange}
-          value={query}
+          value={query ?? ''}
```
j8seangel, line 77: good catch! I'll use that in fishing-map too as I've seen the same warning

applications/vessel-history/src/features/search/search.thunk.ts
```diff
@@ -111,40 +111,35 @@ export type VesselSearchThunk = {
...
-      value: results?.total
+      value: results?.total,
     })
   } else {
     uaEvent({
       category: 'Search Vessel VV',
       action: 'Click Load More',
       label: actualResults.toString(),
-      value: results?.total
+      value: results?.total,
```
j8seangel, line 133: not directly related with the review but aren't we running the pre-commit hook to lint the files in this project? o how did we end up having this in the repo?

applications/vessel-history/src/features/user/user.hooks.ts
```diff
@@ -28,5 +42,16 @@ export const useUser = () => {
     dispatch(logoutUserThunk({ redirectTo: 'home' }))
   }, [dispatch])
 
-  return { loading, logged, user, authorized, login, logout }
+  useEffect(() => {
+    if (!logged && (token || refreshToken || accessToken)) dispatch(fetchUserThunk())
```
j8seangel, line 46: in other apps the token is refreshed automatically when the first request return unauthorised but this way you ensure you are always logged

j8seangel (approved):
great work!

applications/vessel-history/src/features/search/AdvancedSearch.tsx
rdgfuentes: Because if I use directly `fetchResults` then the  click `event` is passed as the first argument when `fetchResults` is expecting the `offset: number` and the app crashes and throws an error, that was my first take yesterday but it took me some time to find what was causing the problem 🤯  🤦

applications/vessel-history/src/features/search/search.thunk.ts
rdgfuentes: Yes we are, but it might be not too strict... I have my IDE set to autoformat on save but maybe some of us does not have eslint plugin properly configured to do so.
We should probably double check and fix the linting rules to  be in sync between the [pre-commit hook settings](https://github.com/GlobalFishingWatch/frontend/blob/develop/package.json#L90-L102) and the [autoformat on save](https://github.com/GlobalFishingWatch/frontend/blob/develop/.vscode/settings.json#L17-L18)

applications/vessel-history/src/features/home/Home.tsx
rdgfuentes: thanks!

applications/vessel-history/src/features/search/search.thunk.ts
j8seangel: yes I have the same configuration and taking a look at the configuration seems to be ok.. lets see if that happens again

PR #981, opened 2021-09-23
clarify no events data

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-223

- [x] added message in activity tab
- [ ] add something in map tab

image


applications/vessel-history/src/features/profile/components/activity/Activity.tsx
```diff
@@ -64,7 +65,7 @@ const Activity: React.FC<ActivityProps> = (props): React.ReactElement => {
     },
     [dispatch, props, setMapCoordinates]
   )
-
+  const hasGFWEvents = useSelector(selectVesselId)
```
rdgfuentes, line 68: I'd change the name of the variable with something more meaningful, maybe something like:
```suggestion
  const isGFWVessel = useSelector(selectVesselId)
```

PR #983, opened 2021-09-24
use npm cdn to fetch locales

j8seangel:
As we need to publish the package for the definition types, makes more sense to have the same source of truth.
This PR replaces GitHub raw files for jsDelivery cdn to consume locales.


rdgfuentes (commented):
Good catch! 👏

PR #999, opened 2021-09-28
Fixed long events description overflow in activity

rdgfuentes:
image


applications/vessel-history/src/features/profile/components/activity/Activity.module.css
```diff
@@ -18,6 +18,10 @@
   padding: 0 var(--space-S);
 }
 
+.event .eventData {
+  overflow: hidden;
```
fertandil87, line 22: add this will looks better https://developer.mozilla.org/es/docs/Web/CSS/text-overflow

PR #1006, opened 2021-09-29
Feature/vv 256

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-256
image


applications/vessel-history/public/locales/source/translations.json
```diff
@@ -236,9 +236,14 @@
       "0": "Not listed",
       "1": "previously listed",
       "2": "currently listed"
+    },
+    "authorizations": {
```
rdgfuentes, line 240: Apparently "authorizations" translation was missing but it was in use. That's why the title of the popup is wrong:
image

Use other key instead, eg: `auths`
```suggestion
    "auths": {
```

fertandil87: wow I didn't see that

PR #1015, opened 2021-10-02
DRY popup context layers rows

nerik:
Try to reduce code repetition on Popup context layer rows to avoid issues such as https://globalfishingwatch.atlassian.net/browse/MAP-594

I haven't touched report and download callbacks, but could we avoid duplicating logic between `ContextLayer` and `UserContextLayer` ?


j8seangel (approved):
Thanks for this! I followed your useHighlightArea hook to unify the `onReportClick` and `onDownloadClick` callbacks Now is much simpler and [I simplify how the component is rendered](https://github.com/GlobalFishingWatch/frontend/commit/08bb8125cdd224c7bb3ea9e237299c429e1eaa10) even more changing only what is different in every layer instead of repeating the use of ContextLayerRow

applications/fishing-map/src/features/datasets/datasets.slice.ts
```diff
@@ -85,10 +86,12 @@ export const fetchDatasetsByIdsThunk = createAsyncThunk(
...
       )
-      const relatedDatasetsIds = initialDatasets.flatMap(
-        (dataset) => dataset.relatedDatasets?.flatMap(({ id }) => id || []) || []
+      const relatedDatasetsIds = uniq(
+        initialDatasets.flatMap(
+          (dataset) => dataset.relatedDatasets?.flatMap(({ id }) => id || []) || []
+        )
       )
-      const uniqRelatedDatasetsIds = without(relatedDatasetsIds, ...ids).join(',')
+      const uniqRelatedDatasetsIds = without(relatedDatasetsIds, ...existingIds).join(',')
```
j8seangel, line 94: Also, this fixes a bug causing an [API 431 error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/431) when too many datasets

PR #1016, opened 2021-10-02
moved parse/stringify workspace to dv-client

nerik:


j8seangel:
Nicely done! we could reuse in other apps now

PR #1087, opened 2021-10-28
Fishing map/4wings compare period graph

satellitestudiodesign:
Screenshot 2021-10-28 at 10 05 59

- [X] Shows a graph of the change compared to baseline period
- [X] Updates the graph when dates or duration change
- [X] Disables before/after mode temporarily
- [X] Removes incorrect time range legend from map


j8seangel:
Thanks for the changes! conflicts fixed and ready to be merged

PR #1099, opened 2021-11-01
VV-323 login not working

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-323

- [X] Moved `useLocalStorage` hook into `react-hooks` package
- [X] Refactored `fishing-map` to consume `useLocalStorage` from `react-hooks` package
- [X] Fixed login not working on `vessel-viewer` in `dev`


applications/vessel-history/src/routes/routes.hook.ts
```diff
@@ -1,10 +1,86 @@
...
+  const { origin, pathname } = window.location
+  return GFWAPI.getLoginUrl(`${origin}${pathname}?${CALLBACK_URL_PARAM}=true`)
+}
+
+export const redirectToLogin = () => {
+  setRedirectUrl()
+  window.location.href = getLoginUrl()
+}
+
+export const useLoginRedirect = () => {
```
j8seangel, line 33: why not having this hook into the package too? as it is almost the same than the fishing one and we need to add support configuration for the [storage keys](https://github.com/GlobalFishingWatch/frontend/pull/1099/files#diff-b2b39c93fd82e82cff2c3cdf34bfc02ce52eb519af2b8dbe838e4b30ec5f9947R16) WDYT ?

j8seangel (approved):

applications/vessel-history/src/routes/routes.hook.ts
rdgfuentes: It makes sense so we reduce the duplicated code. I'll move it

PR #1107, opened 2021-11-03
Layer composer/fix race condition fetch breaks

nerik:
Fixes a race condition when breaks are being fetched.


packages/layer-composer/src/generators/heatmap/heatmap-animated.ts
```diff
@@ -354,15 +365,30 @@ class HeatmapAnimatedGenerator {
...
+    this.breaksCache[cacheKey] = {
+      loading: true,
+      error: false,
+    }
 
     const promise = new Promise((resolve, reject) => {
       breaksPromise.then((breaks) => {
-        this.breaksCache[cacheKey] = { loading: false, error: false, breaks }
-        resolve({ style: this.getStyle(finalConfig), config: finalConfig })
+        // This makes sure we are using the latest config, which may not be the case
```
j8seangel, line 375: Nice catch! this should do the trick

PR #1112, opened 2021-11-03
GA Events - avoid tracking "Display Highlight events" when there is no highlight event

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-300

This PR is more related to a problem in the selectors when comparing objects that not change

### DIFF
--- a/applications/vessel-history/src/features/dataviews/dataviews.slice.ts
+++ b/applications/vessel-history/src/features/dataviews/dataviews.slice.ts
@@ -8,7 +8,6 @@ import { RootState } from 'store'
 export const fetchDataviewsByIdsThunk = createAsyncThunk(
   'dataviews/fetch',
   async (ids: number[], { signal, rejectWithValue, getState }) => {
-    console.log('JJJJJJJJJJJJJJJJJJJJJJJJJJJJ')
     const existingIds = selectIds(getState() as RootState) as string[]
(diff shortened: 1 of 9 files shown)


applications/vessel-history/src/features/dataviews/dataviews.slice.ts
```diff
@@ -62,6 +62,18 @@ export const selectDataviewById = memoize((id: number) =>
...
+    return dataviews
+  }
+)
+
+export const selectDataviewsStatus = createSelector(
+  [selectDataviews],
+  (dataviews) => {
+    return dataviews.status
+  }
+)
```
rdgfuentes, line 77: I'd revert these changes, the problem was not here and having two selectors to consume only one `selectDataviewsStatus` from other places is not ideal.

applications/vessel-history/src/features/dataviews/dataviews.utils.ts
```diff
@@ -102,7 +102,7 @@ export const initializeDataviews = async (dispatch: AppDispatch) => {
   const action = await dispatch(fetchDataviewsByIdsThunk(dataviewIds))
   if (fetchDataviewsByIdsThunk.fulfilled.match(action as any)) {
     dataviews = action.payload as Dataview[]
+    const datasets = getDatasetByDataview(dataviews)
```
rdgfuentes, line 105: 👍

PR #1131, opened 2021-11-12
Update to use yarn 2

j8seangel:
Trying yarn 2 to obtain benefits of yarn-deduplicate
See [The dark side of monorepo and workspaces](https://adevnadia.medium.com/three-simple-tricks-to-speed-up-yarn-install-9cae57d159db)


j8seangel:
Rollback to yarn 1 as finding too many issues

PR #1145, opened 2021-11-17
Css styles are broken after merging nx monorepo

fertandil87:
WIP

some styles are broken because there are changes in the base ui components

this pr is to share ideas


rdgfuentes (approved):
LGTM

PR #1149, opened 2021-11-19
VV-331 Modal popups not working

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-331
- [x] Fixed redirect tp load resources when app is on subdomain and also in path (prod)
- [x] Fixed app title and base urls
- [x] Fixed modal popups not appearing

### DIFF
--- a/apps/vessel-history/nginx.conf
+++ b/apps/vessel-history/nginx.conf
@@ -31,7 +31,7 @@ http {
         listen ${PORT};
         root /usr/share/nginx/www/;
 
-        location ~* ^/map/(.+\..+)$ {
+        location ~* ^/vessel-viewer/(.+\..+)$ {
           alias /usr/share/nginx/www/$1;
         }
 
--- a/apps/vessel-history/pages/_document.tsx
+++ b/apps/vessel-history/pages/_document.tsx
@@ -190,6 +190,7 @@ class MyDocument extends Document {
           </head>
         </Head>
         <body>
+          <div id="root"></div>
           <Main />
           <NextScript />
         </body>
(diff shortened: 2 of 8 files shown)


apps/vessel-history/pages/_document.tsx
```diff
@@ -201,12 +190,13 @@ class MyDocument extends Document {
           </head>
         </Head>
         <body>
+          <div id="root"></div>
```
j8seangel, line 193: I think we can avoid having it here by these two options:
- [use the appSelector prop](https://github.com/GlobalFishingWatch/frontend/blob/develop/apps/fishing-map/features/app/App.tsx#L229) in the Modal ui-component
- update [ui-component default value](https://github.com/GlobalFishingWatch/frontend/blob/develop/libs/ui-components/src/modal/Modal.tsx#L29) as CRA is not used anymore (or keep it as it is and add another fallback value for next apps)

what do you prefer?

apps/vessel-history/nginx.conf
```diff
@@ -31,7 +31,7 @@ http {
         listen ${PORT};
         root /usr/share/nginx/www/;
 
-        location ~* ^/map/(.+\..+)$ {
+        location ~* ^/vessel-viewer/(.+\..+)$ {
```
j8seangel, line 34: ouch, sorry for this! 
my fault copying and pasting from fishing-map

apps/vessel-history/pages/_document.tsx
rdgfuentes: Good catch, thanks! I'll use the `appSelector="__next"` in the Modal ui-component

apps/vessel-history/nginx.conf
rdgfuentes: No problem 👍

PR #1153, opened 2021-11-22
Deploy affected

j8seangel:
Cloud build trigger to decide which apps are affected and needs to be triggered


rdgfuentes:
Awesome! 👍

PR #1155, opened 2021-11-22
vessels pin icon

nerik:
Screenshot 2021-11-23 at 10 19 47


satellitestudiodesign (approved):
cool cool cool
much more consistent with other parts of the popups

PR #1158, opened 2021-11-23
Fishing map/csv to points

j8seangel:
Support uploading .csv files as points dataset converting to geojson on the fly


apps/fishing-map/features/datasets/NewDataset.tsx
```diff
@@ -290,14 +324,15 @@ function NewDataset(): React.ReactElement {
       uaEvent({
         category: `${uaCategory}`,
         action: `Confirm ${uaDatasetType} upload`,
-        label: userTrackGeoJSONFile?.name ?? file.name,
+        label: onTheFlyGeoJSONFile?.name ?? file.name,
       })
+      debugger
```
nerik, line 329: 🐷

apps/fishing-map/features/datasets/NewDataset.tsx
```diff
@@ -108,9 +111,11 @@ function NewDataset(): React.ReactElement {
           file.type === 'application/octet-stream' ||
           file.type === 'multipart/x-zip'
         const isGeojson =
-          (!isZip && file.type === 'application/json') || (!isZip && file.name.includes('.geojson'))
+          !isZip && (file.type === 'application/json' || file.name.includes('.geojson'))
+        const isCSV = !isZip && !isGeojson && file.type === 'text/csv'
```
nerik, line 115: Are we sure we don't ever get `text/plain` here?

libs/data-transforms/src/csv-to-track-segments/guessColumns.ts
```diff
@@ -1,24 +1,44 @@
...
+export const GUESS_COLUMN_DICT: Record<GuessColum, string[]> = {
+  latitude: LatitudeMatches,
+  longitude: LongitudeMatches,
+  timestamp: TimestampMatches,
+}
+
+export const GUESS_COLUMN_NAMES = [Object.entries(GUESS_COLUMN_DICT)]
 
-const guessColumns = (columns: string[] | undefined) => {
+export const guessColum = (col: GuessColum, options: string[]) => {
```
nerik, line 25: Typo: colum -> column
```suggestion
export const guessColumn = (col: GuessColumn, options: string[]) => {
```

libs/data-transforms/src/csv-to-track-segments/guessColumns.ts
```diff
@@ -1,24 +1,44 @@
-const GUESS_COLUMN_NAMES: [string, string[]][] = [
-  ['latitude', ['latitude', 'lat', 'location-lat']],
-  ['longitude', ['longitude', 'lng', 'lon', 'long', 'location-lng']],
-  ['timestamp', ['timestamp', 'time', 'date', 'datetime']],
+export type GuessColum = 'latitude' | 'longitude' | 'timestamp'
```
nerik, line 1: Typo: colum -> column
```suggestion
export type GuessColumn = 'latitude' | 'longitude' | 'timestamp'
```

nerik (changes requested):
Couldn't test the feature because of that error message
Screenshot 2021-11-29 at 13 09 16
:

Translation was incomplete here:
Screenshot 2021-11-29 at 13 02 31

Suggestion: we could have a more future-proof translation mechanism so that we don't have to add a string whenever a format is added, using a placeholder for file formats.
I'm thinking here for instance:
https://github.com/GlobalFishingWatch/frontend/pull/1158/files#diff-0b173523723b8e6317526fddb2aa4ea3bb72f66183ed1c6b76ca1e7d5a479cbdR105-R107
```
"dragFilePlaceholder": "Drag and drop here a file of format: {{allowedFormats}} or click to select it",
```
And here:
https://github.com/GlobalFishingWatch/frontend/pull/1158/files#diff-0b173523723b8e6317526fddb2aa4ea3bb72f66183ed1c6b76ca1e7d5a479cbdR114-R116

```
"onlyFileFormatAllowed": "(Only {{formats}} files are allowed)",
```
"onlyCsvAllowed": "(Only .csv files are allowed)",

j8seangel:
> Couldn't test the feature because of that error message
> Screenshot 2021-11-29 at 13 09 16

Fixed! 

> Suggestion: we could have a more future-proof translation mechanism so that we don't have to add a string whenever a format is added, using a placeholder for file formats.

Indeed, I'll let you know once ready to review again with the changes

apps/fishing-map/features/datasets/NewDataset.tsx
j8seangel: In which use case could it happen?

nerik:
There is an issue IMHO with the points styles at low z levels, ie find the red dots here:

Screenshot 2021-11-30 at 14 47 45

PR #1164, opened 2021-12-02
VV-335 Sw not working workaround

rdgfuentes:
- [x] Implemented `next-pwa` for vessel viewer offline mode
- [x] Adapted precache to include list of generated files (required when building the app from a folder different that the app root)
- [x] Added a `docker-compose.yaml` at the root level to replicate easily the prod environment (with the proxy and the app running in a path other than root `/` )
- [x] Fixed build process, missing `NEXT_PUBLIC_URL` variable
- [x] Cleanup old code for offline mode implementation


apps/vessel-history/next.config.js
```diff
@@ -34,7 +32,62 @@ const nextConfig = {
     }
     return config
   },
+
   // i18n,
+  basePath:
+    process.env.NEXT_PUBLIC_URL || (process.env.NODE_ENV === 'production' ? '/vessel-viewer' : ''),
+  productionBrowserSourceMaps:
```
j8seangel, line 39: Good idea, this way we can see source maps for the staging environment!

docker-compose.yaml
```diff
@@ -0,0 +1,38 @@
+version: '3.9'
```
j8seangel, line 1: Great! now we can test in docker way more easy locally, thanks!
Is there a way to make this available for every app ? otherwise I'll move to the app folder and duplicate in fishing-map when needed

j8seangel (changes requested):
Great work, I love the idea of having a docker-compose 🙇 
However I didn't manage to get it working with the following error:
```
 cannot load certificate "/etc/nginx/ssl/server.crt"
 ```
I run the generate-certificate.sh script and had the same error, any idea?

and is the a way of doing it directly in docker? otherwise I'll include in the README documentation

Thanks!

File: docker-compose.yaml (hunk 2 above)
Comment (rdgfuentes) [reply docker-compose.yaml:1]: I added `map` to the docker-compose, now it can be used too 😉 . The idea is to use this single docker-compose and add inside it all our apps in the monorepo to be proxied like we do in prod.

**Comment by rdgfuentes** [comment]
> Great work, I love the idea of having a docker-compose 🙇 However I didn't manage to get it working with the following error:
> 
> ```
>  cannot load certificate "/etc/nginx/ssl/server.crt"
> ```
> 
> I run the generate-certificate.sh script and had the same error, any idea?
> 
> and is the a way of doing it directly in docker? otherwise I'll include in the README documentation
> 
> Thanks!

@j8seangel The `generate-certificate.sh` was moved to the root folder and the README updated with the steps to this up. Let me know if you find any problem setting this up.

File: apps/fishing-map/.build.env.sample [hunk 3]
```diff
@@ -0,0 +1,16 @@
+BASIC_AUTH_PASS=default
```
j8seangel, line 1: do we need two env files ? why not reuse the .env.sample we already have ?

j8seangel:
> @j8seangel The generate-certificate.sh was moved to the root folder and the README updated with the steps to this up. Let me know if you find any problem setting this up.

@rdgfuentes it worked perfectly, thanks! apart the .env comment it looks good to me

j8seangel:
I'm going to merge so we can test it in dev environment

rdgfuentes:
> @rdgfuentes I've also added this [commit](https://github.com/GlobalFishingWatch/frontend/commit/775adcd90a6b80b7bdb02054ef84c03c5f808c25) to ignore certs in git

Thanks @j8seangel !

apps/fishing-map/.build.env.sample
rdgfuentes: I decided to use two separate .env files to avoid conflicts because one (`.env`) contains the variables that the app needs to run locally; but the other (`.build.env`) has the variables needed in the build processes (`cloudbuild` and `docker-compose`) to build and deploy the app properly.

PR #1185, opened 2021-12-09
Fishing map/fix empty time comparisosn

nerik:


apps/fishing-map/features/analysis/Analysis.tsx
```diff
@@ -226,16 +226,16 @@ function Analysis() {
             disabled={!bounds}
           />
         </div>
+        <div>
+          <p className={styles.placeholder}>
```
j8seangel, line 230: this was hidden on purpose when the button is not visible as it says: `Click the button below if you need a more precise anlysis` and is confusing.
If we always want to show `The data shown above should be taken as an estimate.` then we should split the message

j8seangel (commented):
Take a look at the evolution comment and I'm missing what Matt shared in the feedback document:
```
There’s currently no disclaimer under the ‘before/after’ and ‘period/comparison’ analyses
```

apps/fishing-map/features/analysis/Analysis.tsx
nerik: ok, will do

PR #1203, opened 2021-12-17
Fishing map/viirs radiance filter

j8seangel:
Adds radiance, matched and source filters for viirs layer, test it in  `viirs_match_prototype-public` [workspace](https://fishing-map-fishing-map-viirs-radiance-filter-jzzp2ui3wq-uc.a.run.app/fishing-activity/viirs_match_prototype-public)

image

image

And shows detections by vessel in map popup reusing the `<VesselsTable />` component:

image


apps/fishing-map/features/map/map.slice.ts
```diff
@@ -24,10 +24,11 @@ import { selectUserLogged } from 'features/user/user.slice'
 import { getRelatedDatasetByType, getRelatedDatasetsByType } from 'features/datasets/datasets.utils'
 
 export const MAX_TOOLTIP_LIST = 5
-export const MAX_VESSELS_LOAD = 200
+export const MAX_VESSELS_LOAD = 150
```
j8seangel, line 27: I had some error 431 so I decreased this number a bit

apps/fishing-map/features/map/map.slice.ts
```diff
@@ -192,15 +184,19 @@ export const fetchVesselInfo = async (
   }
 }
 
+export type ActivityProperty = 'hours' | 'detections'
 export const fetchFishingActivityInteractionThunk = createAsyncThunk<
   { vessels: SublayerVessels[] } | undefined,
-  { fishingActivityFeatures: ExtendedFeature[] },
+  { fishingActivityFeatures: ExtendedFeature[]; activityProperty?: ActivityProperty },
```
j8seangel, line 190: added activityProperty to be able to reuse for viirs and fishing

apps/fishing-map/features/map/map.slice.ts
nerik: 👍

apps/fishing-map/features/map/map.hooks.ts
```diff
@@ -54,15 +54,28 @@ import {
...
 export const SUBLAYER_INTERACTION_TYPES_WITH_VESSEL_INTERACTION = [
   'fishing-effort',
   'presence-detail',
+  'viirs-match',
 ]
+export const SUBLAYER_INTERACTION_TYPES_WITH_VIIRS_INTERACTION = ['viirs']
+
+export const getVesselsInfoConfig = (vessels: ExtendedFeatureVessel[]) => {
+  return {
+    vessels: vessels,
```
nerik, line 71:
```suggestion
    vessels,
```

apps/fishing-map/features/map/map.hooks.ts
```diff
@@ -54,15 +54,28 @@ import {
...
 } from './map.slice'
 import useViewport from './map-viewport.hooks'
 import { useMapAndSourcesLoaded } from './map-features.hooks'
 
 export const SUBLAYER_INTERACTION_TYPES_WITH_VESSEL_INTERACTION = [
   'fishing-effort',
   'presence-detail',
+  'viirs-match',
 ]
+export const SUBLAYER_INTERACTION_TYPES_WITH_VIIRS_INTERACTION = ['viirs']
```
nerik, line 67: Do we actually plan to keep that type of interaction? (do we plan to keep VIIRS layer without radiance filter etc?)

apps/fishing-map/features/map/map.slice.ts
```diff
@@ -233,23 +237,23 @@ export const fetchFishingActivityInteractionThunk = createAsyncThunk<
         })
       })
 
-      const topHoursVessels = vesselsBySource
+      const topIncidenceVessels = vesselsBySource
```
nerik, line 240: Took me a while to understand what `incidence` meant in that context. It's generic for 'hours or detections' correct? Then why not `topActivityVessels` or simply `topVessels`?

apps/fishing-map/features/map/popups/VesselsTable.tsx
```diff
@@ -99,70 +134,111 @@ function VesselsTable({
...
+              disabled={!gfwUser}
+              onClick={() => setModalOpen(true)}
+            >
+              + {feature.vesselsInfo.overflowNumber} {t('common.more', 'more')}
+            </button>
+          )}
+        </Fragment>
+      )}
+      {gfwUser && (
+        <Modal
```
nerik, line 225: I'm really not sure about moving the modal logic inside `VesselsTable`. Correct me if I'm wrong, but it means when you instanciate VesselsTable, it instanciates itself within the modal? Which would mean that the modal instanciates another modal for no reason. Is it the reason why the vessels count label at the bottom of the table inside the modal wrongly has a pointer cursor? Maybe worth discussing as I'm not sure what's happening here but it smells like a smell.

nerik: Could be structured this way:
```
FishingRow / Viirs Row
  -> VesselsTablePopupRow
    -> VesselsTable
    -> Modal
      -> VesselsTable  

```

apps/fishing-map/features/workspace/activity/ActivitySchemaFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    const onSliderChange = (rangeSelected) => {
+      if (rangeSelected[0] === range[0] && rangeSelected[1] === range[1]) {
+        onClean(id)
+      } else {
+        const selection = rangeSelected.map((id) => ({
+          id: id.toString(),
+          label: id.toString(),
+        }))
+        onSelect(id, selection)
```
nerik, line 46: I would debounce that.

apps/fishing-map/features/workspace/shared/DatasetSchemaField.tsx
```diff
@@ -14,7 +14,23 @@ type LayerPanelProps = {
 }
 
 function DatasetSchemaField({ dataview, field, label }: LayerPanelProps): React.ReactElement {
-  const valuesSelected = getSchemaFieldsSelectedInDataview(dataview, field)
+  let valuesSelected = getSchemaFieldsSelectedInDataview(dataview, field).sort(
+    (a, b) => a.label - b.label
+  )
+
+  const valuesAreRangeOfNumbers =
+    valuesSelected.length > 1 && valuesSelected.every((value) => Number(value.label))
```
nerik, line 22: Shouldn't you use isNan or typeof here?

nerik (changes requested):
Awesome work ! Love the new Slider component! 
Some general comments along small code-related ones:
- is it worth exploring using a checkbox component for the 'Matched' dropdown? Since we are already using a different UI element for radiance, maybe not too much of a pain?
- the 'source' dropdown shows a 'Null' item
Screenshot 2021-12-23 at 10 59 29
- please add a debounce on the radiance selector as it is spamming the API while dragging
- it seems the map popup's "detection" is not pluralized 
Screenshot 2021-12-23 at 11 12 38

apps/fishing-map/features/map/map.hooks.ts
j8seangel: Nop, this is deprecated and once the prototype is ready for production we'll remove this. Added a comment so we don't forget it

apps/fishing-map/features/workspace/shared/DatasetSchemaField.tsx
j8seangel: I think we are good like this because the method `Number` returns:
- 0 for ''
- NaN for 'asdf'
- 1 for '1'

apps/fishing-map/features/workspace/activity/ActivitySchemaFilter.tsx
j8seangel: Done https://github.com/GlobalFishingWatch/frontend/pull/1203/commits/c3889eb3e9f41907fd0ae7656115c10c27979558

apps/fishing-map/features/map/popups/VesselsTable.tsx
j8seangel: Good catch with the double Modal instances so I've fixed it with this [commit](https://github.com/GlobalFishingWatch/frontend/commit/921775e64c564ce40d6b986a4eca7dbd5fef2406)

Because I think is it much simpler to handle the modal within the own table component that replicating in every PopupRow used

j8seangel:
> is it worth exploring using a checkbox component for the 'Matched' dropdown? Since we are already using a different UI element for radiance, maybe not too much of a pain?
The ideal is to have a multiselect with Matched and NotMatched properties because a checkbox only allows you to select one and would be necessary to have two selected

> the 'source' dropdown shows a 'Null' item
same as above, we need to clean it in the database and remove it from the schema

> it seems the map popup's "detection" is not pluralized
good catch, thanks!

PR #1208, opened 2022-01-04
Allow to add links with label in Crowdin

fertandil87:
images in the ticket https://globalfishingwatch.atlassian.net/browse/VV-337


apps/vessel-history/features/profile/components/GuideAndVideo.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { useCallback, useMemo, useState } from 'react'
+import { event as uaEvent } from 'react-ga'
+import { useTranslation } from 'react-i18next'
+import styles from './Info.module.css'
+
+
+interface GuideAndVideoProps {
+    source?: string
```
rdgfuentes, line 8: Why not using `label` for this prop, it seems more straight forward? What do you think?

apps/vessel-history/features/profile/components/Faq.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useCallback, useMemo, useState } from 'react'
+import { event as uaEvent } from 'react-ga'
+import { useTranslation } from 'react-i18next'
+import styles from './Info.module.css'
+
+
+interface FaqProps {
+    source?: string
```
rdgfuentes, line 8: I'd use `label` for this prop. WDYT?

apps/vessel-history/features/profile/components/GuideAndVideo.tsx
fertandil87: because I was thinking the source of where is clicked

rdgfuentes (approved):
👍

PR #1210, opened 2022-01-06
Fishing map/dynamic ramps

j8seangel:
New useMapDataviewFeatures to read map features using the new event `sourcetilesdata` 

- [x] Dynamic ramps logic
- [x] Analysis, **Bonus:** added blur effect when timeRange for timebar or analysis filter change
- [x] Timebar graph WIP fix next chunk missing graph


apps/fishing-map/features/map/map-sources.hooks.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export const useMapSourceTilesLoadedAtom = () => {
+  // Used it once in Map.tsx the listeners only once
+  const map = useMapInstance()
+  const setSourceTilesLoaded = useSetRecoilState(mapTilesAtom)
+
+  useEffect(() => {
+    if (!map) return
+
+    const onSourceDataLoading = (e: any) => {
```
nerik, line 59: Is there an easy way to type `e` ?

apps/fishing-map/features/map/map-sources.utils.ts
```diff
@@ -0,0 +1,9 @@
+import { ExtendedStyle } from '@globalfishingwatch/layer-composer'
+
+export const getSourceMetadata = (style: ExtendedStyle, id: string) => {
+  const generatorsMetadata = style?.metadata?.generatorsMetadata
+  if (!generatorsMetadata) return null
+
+  const activityHeatmapMetadata = generatorsMetadata[id]
```
nerik, line 7: Rename `activityHeatmapMetadata` as it is meant to be more generic?

apps/fishing-map/features/map/map-sources.hooks.ts
```diff
@@ -0,0 +1,151 @@
...
+          dataviewId: dataview.id,
+          metadata: metadata,
+        }
+      }),
+    [dataviews, style]
+  )
+
+  const dataviewFeatures = useMemo(() => {
+    const dataviewFeature = dataviewMetadata.map(({ dataviewId, metadata }) => {
+      const sourceId = metadata?.timeChunks?.activeSourceId
```
nerik, line 131: As you can see here https://github.com/GlobalFishingWatch/frontend/pull/1210/files#diff-2865ae427da9befe4e55a82d2be9f24a7a506a01c5209013c94fd974056aaac1L65
the timebar used to check all time chunks, not only the active one.
If we only consider the active chunk, we won't see the entirety of the graph in the timebar.

apps/fishing-map/features/timebar/TimebarActivityGraph.utils.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  // let prevMaxFrame: number
+  const { values, maxFrame } = getTimeSeries(features as any, numSublayers, quantizeOffset)
+
+  // TODOTODOTODOTODOTODOTODOTODOTODOTODOTODOTODOTODO
+  // TODO!! ask Erik what does it do  TODOTODOTODOTO!!
+  // TODOTODOTODOTODOTODOTODOTODOTODOTODOTODOTODOTODO
+
+  // const valuesTimeChunkOverlapFramesFiltered = prevMaxFrame
+  //   ? values.filter((frameValues) => frameValues.frame > prevMaxFrame)
```
nerik, line 32: I can explain. Time chunks overlap, ie if you load "1 year" it actually loads 1 year + 3 months (needed for aggregation). So you have 2016 and 2017 as active chunks, you end up with the same values in both chunks for the period january - march 2017. This ensures that when looking at the 2017 chunk, you only use data starting from april 2017.

nerik (changes requested):
Super elegant solution, I love it 👏 
Please have a look at the comments

apps/fishing-map/features/map/map-sources.hooks.ts
j8seangel: yes, there is an `MapDataEvent` that we need to extend with the new sourceId property. I'll include in the fork

PR #1218, opened 2022-01-14
Timebar/centralize derived state

nerik:
Let highlighter reuse derived data (sorted, filtered, clustered) created by charts components without having to explicitely pass them from outside nor recompute them.


j8seangel (approved):
LGTM

PR #1221, opened 2022-01-18
Template app generator

j8seangel:
See [readme](https://github.com/GlobalFishingWatch/frontend/compare/template-app-generator?expand=0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R36) to know how to use the generator


tools/generators/app/files/project.json
```diff
@@ -0,0 +1,103 @@
...
+      },
+      "configurations": {
+        "production": {}
+      }
+    },
+    "start": {
+      "executor": "@nrwl/workspace:run-commands",
+      "options": {
+        "commands": [
+          "nx clean-locales port",
```
rdgfuentes, line 44: Instead of `port` this should be the `<%= name %>`

rdgfuentes (approved):
This is a great start point for bootstrapping new apps @j8seangel! Great work! 👏

tools/generators/app/lib/add-project.ts
```diff
@@ -0,0 +1,132 @@
...
+    // a missing configuration error will be thrown.
+    configurations: {
+      production: {},
+    },
+  }
+
+  targets.start = {
+    executor: '@nrwl/workspace:run-commands',
+    options: {
+      commands: ['nx clean-locales port', 'nx serve i18n-labels', 'nx serve port'],
```
rdgfuentes, line 52: Same here, instead of `port` it should be `${options.projectName}`, right?

fertandil87:
🤓

PR #1233, opened 2022-01-28
Fishing map/map draw polish

j8seangel:
Support deleting points in polygon geometries
image

Disabled when there are only 3 points
image


nerik (approved):
Looks great, works great 👍

PR #1240, opened 2022-01-28
API-190 API Portal minisite

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/API-190


j8seangel:
@rdgfuentes I managed to fix the header confusion.

As you added ui-components into package.json it installed the dependency in node_modules and grab the import from there
image

However I didn't manage to use dynamic import please look at [this commit](https://github.com/GlobalFishingWatch/frontend/pull/1240/commits/89afb295da29b65c1b6d9ece0390d397a825438b) to use the lib from the source instead of the dist version

libs/ui-components/src/index.ts
```diff
@@ -1,7 +1,7 @@
 export * from './types/types'
 export * from './button'
 export * from './color-bar'
-export * from './countryflag'
+// export * from './countryflag'
```
j8seangel, line 4: you found issues with this library, didn't you? Something I need to review why

libs/ui-components/src/slider/Slider.tsx
```diff
@@ -16,11 +16,17 @@ interface SliderProps {
...
 const activeColor =
-  getComputedStyle(document.body).getPropertyValue('---color-primary-blue') ||
-  'rgba(22, 63, 137, 1)'
+  typeof window !== 'undefined'
+    ? getComputedStyle(document.body).getPropertyValue('---color-primary-blue') ||
+      fallbackActiveColor
+    : fallbackActiveColor
 const borderColor =
-  getComputedStyle(document.body).getPropertyValue('--color-border') || 'rgba(22, 63, 137, 0.15)'
+  typeof window !== 'undefined'
```
j8seangel, line 27: thanks for these fixes for the ssr 🙇

j8seangel (approved):
LGTM

libs/ui-components/src/index.ts
rdgfuentes: Yes, it was causing errors when building / starting the next app. We'll have to review it

PR #1264, opened 2022-02-25
adds a keyboard combo to reset workspace

nerik:
Hit 'w' 7 times to remove all url workspace parameters.

### DIFF
--- a/apps/fishing-map/features/app/AppModals.tsx
+++ b/apps/fishing-map/features/app/AppModals.tsx
@@ -1,5 +1,6 @@
 import React, { Fragment, useEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
+import { replace } from 'redux-first-router'
 import { useLocalStorage } from '@globalfishingwatch/react-hooks'
 import { Modal } from '@globalfishingwatch/ui-components'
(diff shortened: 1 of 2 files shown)


apps/fishing-map/features/app/AppModals.tsx
```diff
@@ -93,7 +101,7 @@ const AppModals = () => {
       {gfwUser && (
         <Modal
           appSelector={ROOT_DOM_ELEMENT}
-          title="Big query datasets creation"
+          title="Big query datasets creation 🧠"
```
j8seangel, line 104: 💯

apps/fishing-map/features/app/AppModals.tsx
```diff
@@ -40,13 +40,21 @@ const BigQueryMenuConfig = {
   selectMenuActive: selectBigQueryActive,
 }
 
+const ResetWorkspaceConfig = {
+  key: 'w',
+  onToggle: () => {
+    window.location.replace(window.location.origin)
```
j8seangel, line 46: I think will be safer to use the `replace` method from
```import { replace } from 'redux-first-router'```

j8seangel (commented):
Nice way of reusing the hook, looks good!

PR #1284, opened 2022-03-09
Fishing map/activity layer stats rtk query

j8seangel:
Compare traditional redux-toolkit slice approach against [react-toolkit-query](https://redux-toolkit.js.org/rtk-query/overview) method to fetch layer stats. 
It needs a custom [baseQuery](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#customizing-queries-with-basequery), see [code](https://github.com/GlobalFishingWatch/frontend/compare/fishing-map/activity-layer-stats...fishing-map/activity-layer-stats-rtk-query?expand=1#diff-aaef2cbeabe68e04bfa31d060c708e24c9e5579d76eb85114f6bc8c650413b43R4) but works nice

Apart of having less code boilerplate the main advantage vs the [previous approach](https://github.com/GlobalFishingWatch/frontend/pull/1274) is each request is done individually getting an automatic cache layer effortless and works great because this means changing a filter in one layer doesn't affect the rest.

My concerns so far:
- we'll need to be careful if we start making lot of requests in the components side and lost track
- looks like the reducer stores all the arguments sent to the hook as the query key so we need to be careful with:
image
or creating a [custom cache key](https://github.com/GlobalFishingWatch/frontend/commit/3b17e74c8edc49de976e09a44ba6abca9b461255)
- keep this in mind https://redux-toolkit.js.org/rtk-query/usage/server-side-rendering#server-side-rendering-with-nextjs for the future

Would be nice to keep investigating:
- https://redux-toolkit.js.org/rtk-query/usage/pagination in vessel searchs
- https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching for request depending on others, like fetching the workspace when the user request finishes

### DIFF
--- a/apps/fishing-map/features/workspace/activity/ActivityLayerPanel.tsx
+++ b/apps/fishing-map/features/workspace/activity/ActivityLayerPanel.tsx
@@ -52,7 +52,7 @@ function ActivityLayerPanel({
   const bivariateDataviews = useSelector(selectBivariateDataviews)
   const guestUser = useSelector(isGuestUser)
   const readOnly = useSelector(selectReadOnly)
-  const { data: stats, isLoading } = useGetStatsByDataviewQuery(
+  const { data: stats, isFetching } = useGetStatsByDataviewQuery(
     {
       dataview,
       timerange,
@@ -192,7 +192,7 @@ function ActivityLayerPanel({
           {stats && (
             <div
               className={cx(activityStyles.stats, {
-                [activityStyles.statsLoading]: isLoading,
+                [activityStyles.statsLoading]: isFetching,
               })}
             >
               {showStats ? (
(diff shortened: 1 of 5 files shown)


rdgfuentes:
Great work @j8seangel ! I agree with you, reducing the LoC we need to write to add a new resource consuming services to our apps will not only speed up the development, but also reduce the probability of making mistakes and thusthe time we spend to maintain it later too. As per your concerns this is my input:
> we'll need to be careful if we start making lot of requests in the components side and lost track

In theory if we have many components performing the same request multiple time is a  case that RTK Query should address with its cache layer, however we should be mindful of the [default cache behavior](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#default-cache-behavior) (eg: Requests responses are kept in cache 60 seconds by default if we want it to persist it more time we should configure it with `keepUnusedDataFor` when creating the Api)

> looks like the reducer stores all the arguments sent to the hook as the query key so we need to be careful with:

In this case, I think we should reduce the number of arguments passed to the hook to the minimum possible. That will also ensure we are hitting the cache layer when multiple requests are performed to the same url most of the times.

> keep this in mind https://redux-toolkit.js.org/rtk-query/usage/server-side-rendering#server-side-rendering-with-nextjs for the future

This is great! 👍

PR #1285, opened 2022-03-10
Require user additional fields to manage app tokens API-227

rdgfuentes:


libs/api-client/src/api-client.ts
```diff
@@ -292,7 +292,12 @@ export class GFW_API_CLASS {
               case 'default':
                 return res
               case 'json':
-                return parseJSON(res)
+                return parseJSON(res).catch((e) => {
+                  // When an error occurs while parsing and
+                  // http response is no content, returns an
+                  // empty response instead of an raising error
+                  if (res.status === 204) return
```
j8seangel, line 299: great catch! 👏

apps/api-portal/pages/index.tsx
```diff
@@ -1,24 +1,43 @@
...
+const Home: NextPage = (context) => {
+  const { isUserApplicationsRequiredInfoCompleted } = useGetUserApplications()
+  const router = useRouter()
+
+  useEffect(() => {
+    if (
+      isUserApplicationsRequiredInfoCompleted !== null &&
+      !isUserApplicationsRequiredInfoCompleted
+    ) {
+      router.push('/signup')
```
j8seangel, line 24: would be nice in next frontend meeting to know a bit more about next routing

j8seangel (approved):
LGTM, Looking forward to see differences with rtkquery or similar as there are lot of fetch logic done by hand. thanks!

PR #1286, opened 2022-03-10
Fishing map/bq events

j8seangel:
Support events clusters tiles from BQ datasets

image


apps/fishing-map/features/bigquery/BigQuery.tsx
```diff
@@ -19,6 +24,19 @@ import {
 } from './bigquery.slice'
 import styles from './BigQuery.module.css'
 
+const VisualisationOptions: { id: BigQueryVisualisation; label: string; fieldsHint: string }[] = [
+  {
+    id: '4wings',
+    label: 'Activity (heatmap)',
+    fieldsHint: '(Ensure id, lat, lon, timestamp and value are all present)',
```
nerik, line 31: Do we need to i18n this? Or is the policy not to translate features only accessible to GFW users? We might want to revise this at some point as this might go public at some point and it's easy to forget i18n

apps/fishing-map/features/map/map.hooks.ts
```diff
@@ -240,9 +242,12 @@ export const useClickedEventConnect = () => {
       (f) => f.generatorType === GeneratorType.TileCluster
     )
     if (encounterFeature) {
```
nerik, line 244: Rename to `clusterFeature` ?

nerik (commented):
Code looks good, I've yet to test it in action!

apps/fishing-map/features/bigquery/BigQuery.tsx
j8seangel: I added i18n, it took 5 mins and as you said we are ready in case we release it to the public

PR #1291, opened 2022-03-14
Fishing map/activity filters

j8seangel:
- Remove fixed filters ids to use fieldsAllowed from dataset to render filters
- Support sorting by dataset.config.fieldsOrder
- Support conditional rendering for filters based on [configuration](https://github.com/GlobalFishingWatch/frontend/compare/fishing-map/activity-filters?expand=1#diff-98ce0532eaaa9234e44e8380f15e942ae62e981698a6a4f6e67add3c292c9b85R49)
- Use same workflow for flags with some workarounds to allow sorting

https://user-images.githubusercontent.com/10500650/158187918-89cfeb7e-7984-40a7-9e7c-07d07bd89c6f.mp4

### DIFF
--- a/apps/fishing-map/features/datasets/datasets.utils.ts
+++ b/apps/fishing-map/features/datasets/datasets.utils.ts
@@ -40,6 +40,18 @@ export type SupportedDatasetSchema =
   | 'target_species' // between camelCase or snake_case
   | 'license_category'
 
+type IncomptibleFilter = {
+  id: SupportedDatasetSchema
+  value: any
+  disabled: SupportedDatasetSchema[]
+}
+type IncomptibleFiltersDict = Record<string, IncomptibleFilter[]>
+const INCOMPATIBLE_FILTERS_DICT: IncomptibleFiltersDict = {
+  'public-presence-viirs-match-prototype:v20220112': [
+    { id: 'matched', value: false, disabled: ['source', 'flag', 'shiptype', 'geartype'] },
+  ],
+}
+
 export type SchemaFieldDataview = UrlDataviewInstance | Pick<Dataview, 'config' | 'datasets'>
 
 export const isPrivateDataset = (dataset: Partial<Dataset>) =>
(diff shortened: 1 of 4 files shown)


apps/fishing-map/features/datasets/datasets.utils.ts
```diff
@@ -39,6 +40,18 @@ export type SupportedDatasetSchema =
   | 'target_species' // between camelCase or snake_case
   | 'license_category'
 
+type IncomptibleFilter = {
```
nerik, line 43: Typo
```suggestion
type IncompatibleFilter = {
```

apps/fishing-map/features/datasets/datasets.utils.ts
```diff
@@ -39,6 +40,18 @@ export type SupportedDatasetSchema =
...
+  id: SupportedDatasetSchema
+  value: any
+  disabled: SupportedDatasetSchema[]
+}
+type IncomptibleFiltersDict = Record<string, IncomptibleFilter[]>
```
nerik, line 48:
```suggestion
type IncompatibleFiltersDict = Record<string, IncompatibleFilter[]>
```

apps/fishing-map/features/datasets/datasets.utils.ts
```diff
@@ -39,6 +40,18 @@ export type SupportedDatasetSchema =
...
+const INCOMPATIBLE_FILTERS_DICT: IncomptibleFiltersDict = {
+  'public-presence-viirs-match-prototype:v20220112': [
+    { id: 'matched', value: false, disabled: ['source', 'flag', 'shiptype', 'geartype'] },
```
nerik, line 51: Do we plan to store this config diirectly in the dataset at some point?

nerik (commented):
Cosicas

nerik (commented):
Screenshot 2022-03-15 at 10 28 28

Not a blocker, but I think if you set filters, then set back matching to false, the non-applicable filters should not visible anymore on the TagList

apps/fishing-map/features/datasets/datasets.utils.ts
j8seangel: That is a good point, we could move it to the dataset.configuration object so we make dynamic and not code dependant.
I'd leave as it is for now so it is a bit easier to understand and we can release it now, but something to keep in mind

j8seangel:
> Screenshot 2022-03-15 at 10 28 28
> 
> Not a blocker, but I think if you set filters, then set back matching to false, the non-applicable filters should not visible anymore on the TagList

Thanks! you found a potencial big problem... to fix this now the same logic to identify incompatible selections is used in the multiselect onSelect callback to clean other incompatible filters

PR #1297, opened 2022-03-18
Fishing map/improve range slider

satellitestudiodesign:
- [x] Format the number to fit the space
Screenshot 2022-03-18 at 11 09 25

- [x] Allow having only min and max in dataset
Screenshot 2022-03-18 at 11 10 20

- [x] Calculate step based on range and width of slider
- [x] Avoid sending min a max filters id unneeded


libs/ui-components/yarn.lock
```diff
@@ -0,0 +1,363 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
j8seangel, line 1: why did you need to run `yarn install` in this lib folder? did you have any error? or can we remove this file?

libs/dataviews-client/src/resolve-dataviews.ts
```diff
@@ -327,10 +327,21 @@ export function resolveDataviews(
           const datasetSchema = dataviewInstance.datasets?.find(
             (d) => d.schema?.[filterKey] !== undefined
           )?.schema?.[filterKey]
+
           if (datasetSchema && datasetSchema.type === 'number') {
-            return `${filterKey} >= ${filterValues[0]} AND ${filterKey} <= ${
-              filterValues[filterValues.length - 1]
-            }`
+            const minPossible = Number(datasetSchema.enum[0])
```
j8seangel, line 332: 👏

j8seangel (commented):
LGTM

libs/ui-components/yarn.lock
satellitestudiodesign: I don't remember it, so just a mistake I suppose. I'll delete it

PR #1298, opened 2022-03-18
Fishing map/graticules as reference layer

satellitestudiodesign:
Adds graticules as a new context layer and shows values. Renders thirty, ten, five or single degree lines depending on zoom level.
Screenshot 2022-03-18 at 14 25 03

### DIFF
--- a/apps/fishing-map/data/default-workspaces/workspace.development.ts
+++ b/apps/fishing-map/data/default-workspaces/workspace.development.ts
@@ -13,6 +13,7 @@ import {
   VIIRS_DATAVIEW_ID,
   CLUSTER_ENCOUNTER_EVENTS_DATAVIEW_ID,
   DEFAULT_BASEMAP_DATAVIEW_INSTANCE_ID,
+  GRATICULES_DATAVIEW_ID,
 } from 'data/workspaces'
 import { ENCOUNTER_EVENTS_SOURCE_ID } from 'features/dataviews/dataviews.utils'
 import { WorkspaceState } from 'types'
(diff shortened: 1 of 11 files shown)


apps/fishing-map/data/workspaces.ts
```diff
@@ -32,6 +32,7 @@ export const MPA_RESTRICTED_DATAVIEW_ID = WORKSPACE_ENV === 'development' ? 100
 export const MPA_DATAVIEW_ID = WORKSPACE_ENV === 'development' ? 98 : 176
 export const RFMO_DATAVIEW_ID = WORKSPACE_ENV === 'development' ? 95 : 175
 export const HIGH_SEAS_DATAVIEW_ID = WORKSPACE_ENV === 'development' ? 97 : 174
+export const GRATICULES_DATAVIEW_ID = WORKSPACE_ENV === 'development' ? 246 : undefined
```
j8seangel, line 35: Please don't forget to include it in production too

apps/fishing-map/data/default-workspaces/workspace.development.ts
```diff
@@ -149,6 +150,13 @@ const workspace: Workspace<WorkspaceState> = {
       },
       dataviewId: HIGH_SEAS_DATAVIEW_ID,
     },
+    {
+      id: 'context-layer-graticules',
+      config: {
+        visible: false,
+      },
+      dataviewId: GRATICULES_DATAVIEW_ID,
```
j8seangel, line 158: Would be nice to have it in the [documentation](https://github.com/GlobalFishingWatch/frontend/blob/develop/apps/fishing-map/data/dataviews.md#L15)

j8seangel (commented):
What description should we add? because right now it only says:
image

we also need to include it in the datasets translations, I'll do that part once we agreed in the description

PR #1308, opened 2022-03-22
API Portal. Improved docker image. API-278

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/API-278
- Build and export next application to be used
- Setup and configured a docker image with an alpine nginx server to serve the app
- Verification of required information to use the API moved to its own component


apps/api-portal/nginx.conf
```diff
@@ -0,0 +1,48 @@
...
+    error_log  /var/log/nginx/error.log warn;
+
+    gzip on;
+    gzip_disable "msie6";
+    gzip_vary on;
+    gzip_proxied any;
+    gzip_comp_level 6;
+    gzip_buffers 16 8k;
+    gzip_http_version 1.1;
+    gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
```
fertandil87, line 26: images?

apps/api-portal/nginx.conf
```diff
@@ -0,0 +1,48 @@
+events {
```
j8seangel, line 1: what makes this nginx config different to avoid using the [shared](https://github.com/GlobalFishingWatch/frontend/blob/develop/config/nginx.conf#L2) one?

j8seangel (approved):
LGTM

apps/api-portal/nginx.conf
rdgfuentes: Every app docker image build process is using their own `nginx.conf` (eg: [fishing map](https://github.com/GlobalFishingWatch/frontend/blob/develop/apps/fishing-map/project.json#L19)) that is not aware of other apps paths and they are located in the app's root folder (eg:  [fishing-map nginx.conf](https://github.com/GlobalFishingWatch/frontend/blob/develop/apps/fishing-map/nginx.conf)). The [shared nginx.conf ](https://github.com/GlobalFishingWatch/frontend/blob/develop/config/nginx.conf#L2) is used only in the root [`docker-compose.yaml`](https://github.com/GlobalFishingWatch/frontend/blob/develop/docker-compose.yaml#L75) to configure our local environments the most similar to prod, having the apps running in specific paths and not the root.

PR #1311, opened 2022-03-22
Belize VMS dataset translation sources

rdgfuentes:


libs/i18n-labels/source/datasets.json
```diff
@@ -81,6 +81,79 @@
       "firstTransmissionDate": "firstTransmissionDate"
     }
   },
+  "private-belize-fishing-effort": {
```
j8seangel, line 84: I'd avoid doing this by hand as we are not getting the real dataset configuration.
If you need to do it for development datasets you can run the [crowdin script](https://github.com/GlobalFishingWatch/crowdin-datasets-internationalization) pointing to that environment.
In this case, for example the description in the dataset says: 
```Description pending``` when we already have it.

I've updated the datasets [here](https://github.com/GlobalFishingWatch/datasets-versioning/commit/881ec5b4664d72cb3438c49a28a019007971aa2a) to contain in so next run are updated

j8seangel (approved):

libs/i18n-labels/source/datasets.json
rdgfuentes: Thanks @j8seangel

PR #1314, opened 2022-03-23
Belize VMS Highlight popup. MAP-793

rdgfuentes:


apps/fishing-map/features/workspace/vessels/VesselTranslations.ts
```diff
@@ -43,4 +43,9 @@ const translations = 'used_to_get_dynamic_translations_autogenerated'
 // t('vessel.gearTypes.trap net', 'Trap Net')
 // t('vessel.gearTypes.purse seine', 'Purse Seine')
 
+// Belize gear types
```
j8seangel, line 46: we'll need to run the i18n-parser and merge the translations.json source file before the release. what about doing it separately of this PR?

rdgfuentes: sure, makes sense

PR #1319, opened 2022-03-23
Include feedback form in VV App

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-363


apps/vessel-history/features/home/Home.tsx
```diff
@@ -191,6 +195,16 @@ const Home: React.FC<LoaderProps> = (): React.ReactElement => {
     })
   }, [advancedSearch, query, vesselIds])
 
+  const onFeedbackClick = useCallback(() => {
+    if (userData) {
```
rdgfuentes, line 199: Why are you checking for `userData` here?

apps/vessel-history/features/profile/Profile.tsx
```diff
@@ -244,6 +248,16 @@ const Profile: React.FC = (props): React.ReactElement => {
     return gfwVesselName ? gfwVesselName.value : vessel?.shipname
   }, [vessel])
 
+  const onFeedbackClick = useCallback(() => {
+    if (userData) {
```
rdgfuentes, line 252: Same here, why are you checking for `userData`?

rdgfuentes (changes requested):
I don't see where you are prefilling the fields of the google form. When the user clicks the feedback icon they'll have to complete all of the fields manually, the idea is that the user arrives the feedback from with some of the fields prefilled, as far as I could see we could prefill:
* email
* url
* operating system
* browser

PR #1338, opened 2022-04-01
Fishing map/deck.gl

nerik:
Let's add deck.gl for shits and giggles (for now)

Note: getting the following error when using `yarn install`, I suspect we're going to see more of this in the future:
```
An unexpected error occurred: "expected workspace package to exist for \"@testing-library/react\"".
```

[This](https://github.com/yarnpkg/yarn/issues/7807) seems to be the culprit. Fixed by running `yarn policies set-version 1.18.0` but I'm not sure it's a good idea, let's discuss.


nerik:
Moved to https://github.com/GlobalFishingWatch/frontend/pull/1360 to fix deploy issue with branch name

PR #1346, opened 2022-04-08
revert "hide download track button for non gfw users"

nerik:
Allow again track download for logged in users.


enriquetuya:
@nerik is closing this PR so it doesn't get merged accidentally, we need a more in-depth change here.

PR #1347, opened 2022-04-08
Port labeler: add port ids - final version verified

fertandil87:


apps/port-labeler/types/index.ts
```diff
@@ -65,8 +65,8 @@ export interface PortPosition {
   port_label: string;
   community_label: string;
   point_label: string;
-
   iso3: string;
   distance_from_shore_m: string;
   community_iso3: string;
+  port_iso3: string;
```
rdgfuentes, line 71: Why adding another field for iso3? What's the difference between `port_iso3` and `iso3`?

PR #1349, opened 2022-04-12
Added pelagic geartypes PIPELINE-777

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/PIPELINE-777


apps/fishing-map/features/workspace/vessels/VesselTranslations.ts
```diff
@@ -48,4 +48,11 @@ const translations = 'used_to_get_dynamic_translations_autogenerated'
 // t('vessel.gearTypes.trawler', 'Trawler')
 // t('vessel.gearTypes.purse_seine', 'Purse Seine')
 
+// Pelagic gear types
+// t('vessel.gearTypes.Handline', 'Handline')
```
fertandil87, line 52: is it ok that one comes with upper case, and the rest not?

fertandil87 (approved):

apps/fishing-map/features/workspace/vessels/VesselTranslations.ts
rdgfuentes: it's how it was entered in the pelagic dashboard... maybe we could transform everything to lower case when ingesting the data.

PR #1358, opened 2022-04-20
Offline view - Map view is displaying the track from another vessel

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-375

### DIFF
--- a/apps/vessel-history/features/profile/Profile.tsx
+++ b/apps/vessel-history/features/profile/Profile.tsx
@@ -156,6 +156,7 @@ const Profile: React.FC = (props): React.ReactElement => {
       Array.from(new URLSearchParams(vesselProfileId).keys()).shift() ?? ''
     ).split('_')
 
+    // this is for update the vessel dataview in case that keep cached with the dataview of another vessel
     if (!vesselDataview || 'vessel-' + gfwId !== vesselDataview.id) {
       updateDataview(dataset, gfwId, tmtId)
     }
--- a/apps/vessel-history/features/profile/Profile.tsx
+++ b/apps/vessel-history/features/profile/Profile.tsx
@@ -65,6 +65,60 @@ const Profile: React.FC = (props): React.ReactElement => {
   )
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


apps/vessel-history/features/profile/Profile.tsx
```diff
@@ -90,54 +144,7 @@ const Profile: React.FC = (props): React.ReactElement => {
...
-          if (trackDatasetId) {
-            const eventsRelatedDatasets = getRelatedDatasetsByType(
-              vesselDataset,
-              DatasetTypes.Events
-            )
-
-            const eventsDatasetsId =
-              eventsRelatedDatasets && eventsRelatedDatasets?.length
-                ? eventsRelatedDatasets.map((d) => d.id)
-                : []
```
rdgfuentes, line 147: I think you should call `updateDataview()` here, right?

apps/vessel-history/features/profile/Profile.tsx
```diff
@@ -147,6 +154,7 @@ const Profile: React.FC = (props): React.ReactElement => {
       dispatch(setHighlightedEvent(undefined))
       dispatch(setVoyageTime(undefined))
     }
+    updateDataview()
```
rdgfuentes, line 157: Why not verifying that the `vessel.dataview.id` does not match the current vessel id before calling `updateDataview()`? I think that would be convenient to prevent updating many times the vessel dataview

fertandil87: was my first try just to test and worked so good, and that check that you mention looks that is only to avoid the case a uses exit a profile and join to the same profile, but as all is in the store there is no performance issues or delays @rdgfuentes

fertandil87: you are right, I dont see problems but it is not ok

apps/vessel-history/features/profile/Profile.tsx
fertandil87: no, here we have the same problem

PR #1360, opened 2022-04-21
Fishing map/deckgl

nerik:
- [x] upload PoC tiles to bucket, and use
- [x] use an IconLayer as an MVTLayer sublayer
- [ ] implement time filtering in a custom `TemporalIconLayer`
- [ ] Style


j8seangel:
As we are planning to replace maplibre with deck.gl entirely we are closing this in favour of https://github.com/GlobalFishingWatch/frontend/pull/1612

PR #1368, opened 2022-04-27
Coastal Fisheries - missing data in timebar  and others

fertandil87:
https://globalfishingwatch.atlassian.net/browse/MAP-784


j8seangel:
@fertandil87 these changes are not always working... I tried to fix what we already had and seems to be working [here](https://github.com/GlobalFishingWatch/frontend/pull/1370) please take a look

PR #1370, opened 2022-04-27
fix missing features filtered by viewport in timebar

j8seangel:
Found a use case that https://github.com/GlobalFishingWatch/frontend/pull/1368 doesn't cover so this PR fixes the filter by viewport issue using the same approach than before but moving only the needed features.


j8seangel:
Merging to solve the current issue with the fewer changes possible, let's review https://github.com/GlobalFishingWatch/frontend/pull/1368 and compare later

PR #1371, opened 2022-04-27
Coastal Fisheries - missing data in timebar and others

fertandil87:
https://globalfishingwatch.atlassian.net/browse/MAP-784


nerik:
@fertandil87 I don't see any problem in merging this, always good to have failsafes, but I don't think this is a fix to MAP-784. The problem sometimes happens, sometimes not, with or without this fix. I'll be investigating further.

PR #1374, opened 2022-04-28
Fishing map/layer icons

j8seangel:
Show icon by layer type in:
- [x] Map tooltips
- [x] Timebar tooltips
image
image
image
image
image
image


enriquetuya:
@j8seangel @satellitestudiodesign loved how this looks, improves clarity a lot. 

Though it is still a bit unclear when you have both encounters for a vessel and the encounters layer. But I need to check again once the tracks are loaded

libs/ui-components/src/icon/Icon.tsx
```diff
@@ -16,13 +16,16 @@ import { ReactComponent as CategoryMarineManager } from '../assets/icons/categor
 import { ReactComponent as CategoryNews } from '../assets/icons/category-news.svg'
 import { ReactComponent as CategoryReports } from '../assets/icons/category-reports.svg'
 import { ReactComponent as Close } from '../assets/icons/close.svg'
+import { ReactComponent as Clusters } from '../assets/icons/clusters.svg'
```
nerik, line 19: IMHO this file deserves to be auto generated from an `ls assets/icons/`. Just an idea, maybe overkill.

nerik (approved):
This is looking really good, thanks 🥲
Let me know what you think about autogenerating `Icon.tsx`, otherwise looks perfect, will merge.

libs/ui-components/src/icon/Icon.tsx
j8seangel: I like the idea! I would use something like [this](https://github.com/cycloidio/import-index-generator/blob/master/main.js#L44) as base to do it

PR #1381, opened 2022-04-28
UX Improvements. VV-429

rdgfuentes:
- Autofocus on vessel name when landing on the search page
- Added option to clear all fields on vessel search 
- Submit on enter key press
- Fixed clumpy width on vessel home when viewing offline vessels vs vessel's search results
- Enhanced Button and IconButton component to accept `htmlType` property that will set the `type` property of the actual html button being rendered so we can define it to `submit` | `button` | `reset` to use them in forms.
- Enhanced InputDate, InputText, Select, MultiSelect and Textarea components to accept `id` property instead of using always the `label` as `id` (it was causing the vessel name field to show in the autocomplete list names and emails)


j8seangel (approved):
Great workaround for the htmlType! thanks for adding the id and fixing the label, great work 🙇

PR #1386, opened 2022-05-04
sort highlighted events

satellitestudiodesign:
before:
Screenshot 2022-05-04 at 11 15 12

after:
Screenshot 2022-05-04 at 11 15 29


nerik (approved):
Looking good, thanks, added logic for non-fishing events.

PR #1405, opened 2022-05-11
Fix/react map gl build

j8seangel:
Fix apps build after react-map-gl updated


j8seangel:
Merging to fix the builds, but @fertandil87 would be nice if you can take a look at the react-map-gl migration on the port-labeller

PR #1411, opened 2022-05-18
use mapbox url to fetch the resource

j8seangel:
Using mapbox to fetch the geojson and the `useMapSourceTilesLoaded` to render the loading state for the footprint layer

Pros: it simplifies the workflow and remove a lot of code
Cons: the layer remains on the map until the new geojson is loaded


nerik (commented):
It looks much cleaner IMHO. I've added a proposal to show the polygon layeer's loading status, let me know what you think.

PR #1412, opened 2022-05-19
Vessel groups

nerik:


nerik:
Test CSV
[VesselGroupsTest.csv](https://github.com/GlobalFishingWatch/frontend/files/8992683/VesselGroupsTest.csv)

libs/api-client/src/api-client.ts
```diff
@@ -280,26 +280,28 @@ export class GFW_API_CLASS {
...
+              email: process.env.REACT_APP_LOCAL_API_USER_EMAIL,
+            }),
+          }),
+          Authorization: `Bearer ${this.getToken()}`,
+        }
         const fetchUrl = isUrlAbsolute(url)
           ? url
           : this.baseUrl + (dataset ? `/datasets/${this.dataset}` : '') + url
         const data = await fetch(fetchUrl, {
+          redirect: 'follow',
```
j8seangel, line 300: what is this? and why is it needed as a fixed param?

libs/ui-components/src/multi-select/MultiSelect.tsx
```diff
@@ -279,18 +281,21 @@ export function MultiSelect(props: MultiSelectProps) {
...
-                  : highlight || isSelected
+                  : (highlight || isSelected) && !item.disableSelection
                   ? 'tick'
                   : ('' as IconType)
               return (
                 <Tooltip key={item.id} content={item.tooltip} placement="top-start">
                   <li
                     className={cx(styles.optionItem, {
                       [styles.highlight]: highlight,
+                      [item.className || '']: item.className,
```
j8seangel, line 298: Don't understand this because the item.className will be only added when it exists
the fallback is useful when is used as:
`cx(styles[item.className])`
so this can be simplified as:
`[item.className]: item.className`

apps/fishing-map/features/vesselGroup/vessel-groups.selectors.ts
```diff
@@ -0,0 +1,13 @@
...
+import { isAdvancedSearchAllowed } from 'features/search/search.selectors'
+import { isGFWUser } from 'features/user/user.slice'
+import { RootState } from 'store'
+
+// export const selectVesselGroups = (state: RootState) => state.vesselGroups.data
+
+export const selectVessselGroupsAllowed = createSelector(
+  [isAdvancedSearchAllowed, isGFWUser],
+  (advancedSearchAllowed, gfwUser) => {
+    return gfwUser || advancedSearchAllowed
```
j8seangel, line 11: what is the relationship between vessel-groups and advanced search ?

apps/fishing-map/features/common/FileDropzone.tsx
```diff
@@ -7,20 +7,20 @@ import { ReactComponent as FilesCsvIcon } from 'assets/icons/file-csv.svg'
 import { ReactComponent as FilesJsonIcon } from 'assets/icons/file-json.svg'
 import { ReactComponent as FileZipIcon } from 'assets/icons/file-zip.svg'
 import { joinTranslatedList } from 'features/i18n/utils'
-import styles from './NewDataset.module.css'
+import styles from './FileDropzone.module.css'
```
j8seangel, line 10: Great!!

j8seangel (changes requested):
Some comments related to the code in the review but I also created a new PR with the changes needed so it is easier to track, please take a look at:

https://github.com/GlobalFishingWatch/frontend/pull/1501

libs/api-client/src/api-client.ts
nerik: I was having issues with the API and this was one intent to make it work, removing.

libs/ui-components/src/multi-select/MultiSelect.tsx
nerik: Because `className` is optional in the component props, the TS compiler will complain that `A computed property name must be of type 'string', 'number', 'symbol', or 'any'.` This is IMHO the least bad solution, others include `[item.className as any]: item.className` or `[!item.className]: item.className`

apps/fishing-map/features/vesselGroup/vessel-groups.selectors.ts
j8seangel: > I believe a GFW user always has the right to access advanced search, so this might be redundant

maybe but I think there is a good idea to do it

PR #1413, opened 2022-05-19
simplify generator and hook

j8seangel:
- I don't think we need another param in the generator config for `loaded` because we can already handle the opacity at the [app level](https://github.com/GlobalFishingWatch/frontend/pull/1413/files#diff-36b02ca52df9696dd5bf7a310036c24a3f7a6890c3fee29e951f42cd99bb3af8R39)
- I also think adding an extra parameter for the hook is unnecessary because adds complexity and we can read the source status by id using the existing `useMapSourceTiles` [hook](https://github.com/GlobalFishingWatch/frontend/pull/1413/files#diff-224a0a8337cd448587a358158edf27b37fe7a1b5d58f6f4e1f8000af706c3ff2R81)

### DIFF
--- a/apps/fishing-map/features/map/map.hooks.ts
+++ b/apps/fishing-map/features/map/map.hooks.ts
@@ -32,7 +32,7 @@ import {
   selectShowTimeComparison,
   selectTimeComparisonValues,
 } from 'features/analysis/analysis.selectors'
-import { useMapClusterTilesLoaded, useMapSourceTilesLoaded } from 'features/map/map-sources.hooks'
+import { useMapClusterTilesLoaded, useMapSourceTiles } from 'features/map/map-sources.hooks'
 import { ENCOUNTER_EVENTS_SOURCE_ID } from 'features/dataviews/dataviews.utils'
 import { useAppDispatch } from 'features/app/app.hooks'
 import {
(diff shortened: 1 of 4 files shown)


apps/fishing-map/features/map/map.hooks.ts
```diff
@@ -78,21 +78,18 @@ export const useGeneratorsConnect = () => {
...
         return {
           ...generatorConfig,
-          loaded: layersLoaded[i],
+          opacity: sourceTilesLoaded[generatorConfig.id]?.loaded ? 1 : 0,
         }
       }
       return generatorConfig
     })
-  }, [generatorsConfig, layersLoadedSerialized])
+  }, [generatorsConfig])
```
nerik, line 92: Are you sure you don't need to add `sourceTilesLoaded` in the dependency array?

nerik (approved):
Looking good! I forgot we had `opacity` as a generic generator prop, much better!

apps/fishing-map/features/map/map.hooks.ts
j8seangel: This is what the linter says... and I've tested and it works as expected 🤷‍♂️
image

PR #1415, opened 2022-05-20
Fix port labeler map

j8seangel:


fertandil87:
@j8seangel the map is working, but I have other issues, but I think I can fix them. I will take a look

PR #1418, opened 2022-05-24
Feature/error states

nerik:
Aggressively display errors when API fails:
- fixed error status in sidebar if one of the track chunk of a set fails to load
- added an error message in the Timebar when one of the activity tiles fails to load:
Screenshot 2022-05-24 at 10 47 56
- added same error message when one of the track chunk of a set fails to load, or when one of the displayed events resource fails to load


j8seangel (approved):
💯

PR #1419, opened 2022-05-24
Limit zoom level on animated heatmap layers. MAP-797

rdgfuentes:
- Enable maxZoom on heatmap animated dataviews so that 4wing tiles beyond that zoom level are not requested and loaded.

https://user-images.githubusercontent.com/2205831/170121175-8f3bbe51-a029-4f3b-b56d-4099b0156661.mp4

### DIFF
--- a/libs/dataviews-client/src/resolve-dataviews-generators.ts
+++ b/libs/dataviews-client/src/resolve-dataviews-generators.ts
@@ -443,8 +443,8 @@ export function getDataviewsGeneratorConfigs(
     })
 
     const maxZoomLevels = dataviews
-      ?.filter(({ config }) => config && config?.maxZoom)
-      .flatMap(({ config }) => config?.maxZoom)
+      ?.filter(({ config }) => config && config?.maxZoom !== undefined)
+      .flatMap(({ config }) => config?.maxZoom as number)
     const mergedActivityDataview = {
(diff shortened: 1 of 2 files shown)


libs/dataviews-client/src/resolve-dataviews-generators.ts
```diff
@@ -442,12 +442,22 @@ export function getDataviewsGeneratorConfigs(
...
       config: {
         type: GeneratorType.HeatmapAnimated,
         sublayers: activitySublayers,
         mode: heatmapAnimatedMode,
+        // if any of the activity dataviews has a max zoom level defined
+        // apply the minimum max zoom level (the most restrictive approach)
+        ...((maxZoomLevels &&
+          maxZoomLevels.length > 0 && {
+            maxZoom: Math.min(...maxZoomLevels),
+          }) ||
```
j8seangel, line 459: do we need the or condition ?
I think if none of `maxZoomLevels && maxZoomLevels.length > 0` it won't add anything into the config anyway

j8seangel (commented):
WDTY about adding the `maxZoom` property into the [dataview config types definitino](https://github.com/GlobalFishingWatch/frontend/blob/develop/libs/api-types/src/dataviews.ts#L5) so at least we know it is an option?

PR #1429, opened 2022-05-25
Port Labeler // User wants to be able to see the anchorage-id in order to find quickly an specific anchorage

fertandil87:
https://globalfishingwatch.atlassian.net/browse/LABELER-119


rdgfuentes:
👍

PR #1433, opened 2022-05-26
We want IDS to be unique to avoid collisions on ports and subareas

fertandil87:
https://globalfishingwatch.atlassian.net/browse/LABELER-121


rdgfuentes (commented):
LGTM 👍

PR #1462, opened 2022-06-03
Encounters in foreign EEZs risk indicator. VV-386

rdgfuentes:


apps/vessel-history/features/risk-summary/risk-summary.tsx
```diff
@@ -53,11 +54,19 @@ export function RiskSummary(props: RiskSummaryProps) {
     },
     [highlightEvent, props, setMapCoordinates, viewport.zoom]
   )
+  const hasEncountersInMPAs = useMemo(() => encountersInMPA.length > 0, [encountersInMPA.length])
+  const hasEncountersInForeignEEZs = useMemo(
+    () => encountersInForeignEEZ.length > 0,
+    [encountersInForeignEEZ.length]
+  )
+  const hasFishingInMPAs = useMemo(() => fishingInMPA.length > 0, [fishingInMPA.length])
```
j8seangel, line 62: Do we really need a memo for storing a boolean? 
I found [this article](https://canimerge.com/should-you-really-use-usememo-in-react-lets-find-out/) and I don't think worth it for the complexity of the operation

j8seangel (commented):
A bit out of context, but hope the comments help!

apps/vessel-history/data/eezInfo.ts
```diff
@@ -0,0 +1,1978 @@
+export enum EEZTypes {
```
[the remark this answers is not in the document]
rdgfuentes: I think we can not because they are not including the eez type and the sovereign countris for each eez region.

apps/vessel-history/features/risk-summary/risk-summary.tsx
rdgfuentes: Apparently, we don't need it , good catch!

PR #1475, opened 2022-06-10
Fishing map/allow reordering context layers

satellitestudiodesign:


j8seangel:
https://github.com/GlobalFishingWatch/frontend/pull/1477

PR #1487, opened 2022-06-20
Fishing map/add protected seas layer

satellitestudiodesign:
Screenshot 2022-06-20 at 19 28 01

- [X] Add layer
- [X] Link to protectedSeas for each polygon
- [ ] Create dataview in prod and add to default workspace?


nerik (commented):
Looking good, thanks!

I think it would make sense and declutter to split the layers into 3 layers depending on the protection level. 

Also, I'm in favour of swapping the (i) icon to an "external link" one as the user might expect a modal to open rather than something completely breaking context.
Screenshot 2022-06-21 at 11 34 13

Code-wise, the large branching section in ContextLayers.tsx (multiple else if) could benefit from a more elegant organisation. Either use a switch, or a dictionary associating `generatorContextLayer` with output (label, id, linkHref). But definitely optional, do it if you fancy doing it.

Also, I removed this:
Screenshot 2022-06-21 at 11 56 46

PR #1488, opened 2022-06-21
Fourwings dist fix

j8seangel:
Depends on https://github.com/GlobalFishingWatch/maplibre-gl-js/pull/6

### DIFF
--- a/libs/layer-composer/package.json
+++ b/libs/layer-composer/package.json
@@ -12,7 +12,7 @@
     "url": "git+https://github.com/GlobalFishingWatch/frontend.git"
   },
   "dependencies": {
-    "@globalfishingwatch/maplibre-gl": "^2.1.9-gfw.23",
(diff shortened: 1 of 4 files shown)


libs/layer-composer/src/generators/heatmap/heatmap-animated.ts
```diff
@@ -108,6 +113,39 @@ const serializeBaseSourceParams = (params: TileAggregationSourceParams) => {
...
 
+const getFinalurl = (
+  config: GlobalHeatmapAnimatedGeneratorConfig,
+  params: TileAggregationSourceParamsSerialized
+) => {
+  const { datasets, filters, 'vessel-groups': vesselGroups, ...rest } = params
+  const finalUrlParams = {
+    ...rest,
+    format: 'intArray',
+    'temporal-aggregation': params.singleFrame === 'true',
```
nerik, line 124: I think this is always false, params.singleFrame === 'true' only in non animated heatmap

PR #1489, opened 2022-06-21
SPIKE: Improve how to display search results when there are several o…

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-191


apps/vessel-history/features/home/Home.tsx
```diff
@@ -96,26 +96,37 @@ const Home: React.FC<LoaderProps> = (): React.ReactElement => {
...
   )
 
   const onSeeVesselClick = useCallback(() => {
-    const selectedVessel = vessels[selectedVessels[0]]
+    const parsedSelectedVessels = getListOfSelectedVessels()
+    const selectedVessel = parsedSelectedVessels[0]
     if (selectedVessel) openVesselProfile(selectedVessel)
   }, [openVesselProfile, selectedVessels, vessels])
 
   const onMergeVesselClick = useCallback(() => {
```
rdgfuentes, line 120: This is never executed now when the user only selects an auto-merged vessel because on lines 316:327 of this same file we are displaying `SEE VESSEL` button when there is only one vessel selected (not MERGE SELECTED VESSELS button). I'd leave only the `SEE VESSEL` button there and inside the onSeeVesselClick open the profile of a single vessel when only one vessel without related vessels is selected, but if multiple vessels are selected or one vessel with related vessels then open the profile for all of them. Take into account that browsers do have a maximum length for urls (https://www.geeksforgeeks.org/maximum-length-of-a-url-in-different-browsers/)

rdgfuentes (changes requested):
I like the approach. Added some comments inline for review. I was not able to verify the activity and map (tracks) merged because of the issue with the SEE VESSEL button I mentioned there.

apps/vessel-history/features/home/Home.tsx
fertandil87: no, both buttons are used, so this function is used too

apps/vessel-history/features/search/search.utils.ts
```diff
@@ -0,0 +1,37 @@
...
+        mergedVessels[existingIndex] = getMergedVessel(vessel, mergedVessels[existingIndex],)
+        // add data to main result
+      } else {
+        const newVessel = getMergedVessel(vessel)
+        mergedVessels.push(newVessel)
+      }
+    } else {
+      const newVessel = getMergedVessel(vessel)
+      mergedVessels.push(newVessel)
+    }
```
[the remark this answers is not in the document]
fertandil87: I understand you but you removed the findIndex and you keep using, and if I add it I am doing two searchs to remove a IF, I will modify the push to one line

PR #1501, opened 2022-06-28
Feature/vessel groups review

j8seangel:
- [x] Rename vesselGroup to vessel-group to match dataset fields allowed and remove mocked dataset
- [ ] Fix why there are less vessel groups in user panel, than the filter options
- [x] Add option in user panel to create a vessel group in the same way we have for datasets or workspaces
- [x] Remove all the console.logs and commented code
- [x] ~How can I have a vessel group of 2 ids and have 5 different results in the interaction?~ the vessel-group param is missing in the 4wings interaction request
- [x] Include vessel-group in stats requests
- [x] Include vessel-group in reports and downloads
- [ ] When not logged in user groups will be always empty, should we show at least something to know the feature is there?
- [ ] When creating a vessel group from the filter I'd add the recently created group to the current filters on save
- [ ] The vessel-group list fetch is done twice
- [ ] Group vessel results by id field in search result list

image


nerik:
> Fix why there are less vessel groups in user panel, than the filter options

This is because some vessel groups have been created with another ownerId, and in user groups they are filtered by the current user's ownerId

PR #1527, opened 2022-07-07
Load risk indicators from API and match to rendered events. VV-403

rdgfuentes:


libs/dataviews-client/src/resources/resources-slice.ts
```diff
@@ -40,7 +40,7 @@ export const getTracksChunkSetId = (datasetConfig: DataviewDatasetConfig) => {
 const parseEvent = (event: ApiEvent, eventKey: string): ApiEvent => {
   return {
     ...event,
-    id: eventKey,
+    key: eventKey,
```
rdgfuentes, line 43: @j8seangel I had to change this so that the event id is not overwritten on resources fetching and we can use it to match the event ids we receive from the indicators endpoint to match the events we are using on activity tab.

PR #1532, opened 2022-07-12
Fixed VV is not loading any data in DEV using API V2 because of required pagination. VV-404

rdgfuentes:


apps/vessel-history/features/dataviews/dataviews.slice.ts
```diff
@@ -1,19 +1,25 @@
 import { createAsyncThunk, createSelector } from '@reduxjs/toolkit'
 import { memoize, uniqBy } from 'lodash'
+import { stringify } from 'qs'
```
fertandil87, line 3: is this different from JSON.stringify? @rdgfuentes

fertandil87 (approved):

apps/vessel-history/features/dataviews/dataviews.slice.ts
rdgfuentes: Yes @fertandil87, it's the `stringify` function from `qs` npm package (https://github.com/ljharb/qs#stringifying)

PR #1542, opened 2022-07-13
Migrate User Management Tool to API V2 // ENG-378

fertandil87:
https://globalfishingwatch.atlassian.net/browse/ENG-378

### DIFF
--- a/libs/api-client/src/api-client-v2.ts
+++ b/libs/api-client/src/api-client-v2.ts
@@ -1,8 +1,16 @@
 import { FetchOptions, GFW_API_CLASS } from './api-client'
 
 export const LAST_API_VERSION = process.env.NEXT_PUBLIC_LAST_API_VERSION || 'v2'
+const DEBUG_API_REQUESTS: boolean = process.env.NEXT_PUBLIC_DEBUG_API_REQUESTS === 'true'
 
 export class GFW_API_CLASS_V2 extends GFW_API_CLASS {
+
+  constructor({
+    debug = DEBUG_API_REQUESTS,
+  } = {}) {
+    super({ debug })
+
+  }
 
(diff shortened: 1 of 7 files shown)


libs/api-client/src/api-client-v2.ts
```diff
@@ -0,0 +1,18 @@
+import { FetchOptions, GFW_API_CLASS } from './api-client'
```
rdgfuentes, line 1: Nice approach, super clean and extensible! 👏

rdgfuentes (approved):

libs/api-client/src/api-client-v2.ts
j8seangel: I think we could do the same without the need of any extra code to maintain.
We set the `API_GATEWAY` [env var here](https://github.com/GlobalFishingWatch/frontend/blob/develop/libs/api-client/src/api-client.ts#L10) with the `/v2` suffix as then it is used as baseUrl in the [_internalFetch](https://github.com/GlobalFishingWatch/frontend/blob/develop/libs/api-client/src/api-client.ts#L298)

Something I propose is to chat and refactor the library for the use cases we have now, because for example we have this [legacy dataset config](https://github.com/GlobalFishingWatch/frontend/blob/develop/libs/api-client/src/api-client.ts#L160) from v0 we could remove

cc @fertandil87 @rdgfuentes

rdgfuentes: @j8seangel We talked about that with @fertandil87 but I think it would be better to do it this way until we have all of the apps and endpoints migrated to v2 in dev and prod. We have some endpoints we use that do not have any prefix in the path. Let's chat about this and the refactor you propose.

PR #1546, opened 2022-07-14
Improve new api client & migrate fishing-map and VV

fertandil87:


libs/api-client/src/api-client.ts
```diff
@@ -153,10 +161,15 @@ export class GFW_API_CLASS {
...
+    this.apiVersion = version
+  }
+
   setConfig(config: LibConfig) {
-    const { debug = this.debug, baseUrl = this.baseUrl, dataset = this.dataset } = config
+    const { debug = this.debug, baseUrl = this.baseUrl, dataset = this.dataset, version = this.apiVersion } = config
     this.debug = debug
     this.baseUrl = baseUrl
+    this.apiVersion = version
     this.dataset = dataset
```
j8seangel, line 173: could we get rid of the dataset property? I think it is not used anymore so we can also simplify the url generation [here](https://github.com/GlobalFishingWatch/frontend/blob/migrate-new-api-client/libs/api-client/src/api-client.ts#L319)

libs/api-client/src/api-client.ts
```diff
@@ -229,8 +242,16 @@ export class GFW_API_CLASS {
     return
   }
 
+  generateUrl(url: string, version?: ApiVersion): string {
+    if (isUrlAbsolute(url)) {
+      return url
+    }
```
j8seangel, line 248: what about checking if the url already starts with a ApiVersion value with something like:
```
if (url.startsWith('v1') || url.startsWith('v2')) {
  return url
}
```
so we avoid [cleaning it by hand](https://github.com/GlobalFishingWatch/frontend/pull/1546/files#diff-c21edc9c2057bed8019b6df684a7685168b615f521d45c9f6ffc1bc8e62f5a41R131) in the case where the urls are autogenerated by the datasets endpoints definition

apps/fishing-map/queries/stats-api.ts
```diff
@@ -48,7 +47,8 @@ export const dataviewStatsApi = createApi({
   reducerPath: 'dataviewStatsApi',
   serializeQueryArgs: serializeStatsDataviewKey,
   baseQuery: gfwBaseQuery({
-    baseUrl: `/${API_VERSION}/4wings/stats`,
+    // TODO 1?
```
j8seangel, line 50: what does `1` mean?

apps/fishing-map/features/vessel-groups/vessel-groups.slice.ts
```diff
@@ -128,7 +128,7 @@ export const searchVesselGroupsVesselsThunk = createAsyncThunk(
             message: 'Missing search url',
           })
         }
-        const searchResults = await GFWAPI.fetch<APIPagination<VesselSearch>>(url, { signal })
+        const searchResults = await GFWAPI.fetch<APIPagination<VesselSearch>>(url, { signal, version: '' })
```
j8seangel, line 131: please see comment below, I think we could avoid the `version:''` where the urls are generated by `resolveEndpoint`

j8seangel (dismissed):
Please take a look at the comments but I think we are very close to simplify a lot the use of the library. thanks!

apps/fishing-map/queries/stats-api.ts
fertandil87: nothing, I forget it

libs/api-client/src/api-client.ts
```diff
@@ -14,6 +14,9 @@ const API_GATEWAY =
   'https://gateway.api.dev.globalfishingwatch.org'
 export const USER_TOKEN_STORAGE_KEY = 'GFW_API_USER_TOKEN'
 export const USER_REFRESH_TOKEN_STORAGE_KEY = 'GFW_API_USER_REFRESH_TOKEN'
+export const LAST_API_VERSION = process.env.NEXT_PUBLIC_LAST_API_VERSION || 'v2'
```
rdgfuentes, line 17: I'd use `API_VERSION` and `NEXT_PUBLIC_API_VERSION` as variable names, we don't need the `LAST_`

libs/api-client/src/api-client.ts
rdgfuentes: It makes sense to me, I'd check it against the version prefix that would be added
```typescript
    const apiVersion = version ?? this.apiVersion
    const prefix = apiVersion ? `/${apiVersion}` : ''
    if (url.startsWith(`${prefix}/`)) {
      return url
    }
    return `${prefix}${url}`
```

apps/fishing-map/features/user/UserInfo.tsx
```diff
@@ -2,12 +2,13 @@ import { useCallback, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { Trans, useTranslation } from 'react-i18next'
 import { Button, Spinner } from '@globalfishingwatch/ui-components'
+import { GUEST_USER_TYPE } from '@globalfishingwatch/api-client'
```
j8seangel, line 5: I like this! does it makes sense to move [this other ones](https://github.com/GlobalFishingWatch/frontend/pull/1546/files#diff-dd7c89a932640c2a94ce168c7f83787f30219bd4e6493c5e318a9f3966d92f7bR25) too?

apps/fishing-map/features/user/user.slice.ts
```diff
@@ -37,19 +37,11 @@ export const PRIVATE_SUPPORTED_GROUPS = [
   'Belize',
 ]
 
-export const fetchGuestUser = async () => {
```
j8seangel, line 40: 👍 👏

libs/api-client/src/api-client.ts
```diff
@@ -31,26 +38,26 @@ interface UserTokens {
...
-  baseUrl?: string
-  dataset?: string
-}
-
 interface LoginParams {
   accessToken?: string | null
   refreshToken?: string | null
 }
-
+export type ApiVersion = '' | 'v1' | 'v2'
```
j8seangel, line 45: With [this change](https://github.com/GlobalFishingWatch/frontend/pull/1546/files#diff-096d4e97610c7a4951df4066d454cc29cdbc9a060252a6bd9b9ee18483f2bb99R249) and knowing know that every endpoint is now prefixed by the version.

does it makes sense to keep the `''` ?

libs/api-types/src/user.ts
```diff
@@ -62,3 +62,15 @@ export interface UserData {
   pullingDataOtherAPIS?: string
   apiTerms?: string
 }
+
+export interface UserFetchResponse {
```
j8seangel, line 66: We already have a [paginated type with a Generic](https://github.com/GlobalFishingWatch/frontend/blob/develop/libs/api-types/src/pagination.ts#L1) to define the entries types, I think I'd remove this to avoid duplications

apps/fishing-map/features/user/UserInfo.tsx
fertandil87: I moved this because I needed to update the API url, maybe we can the others in another PR

libs/api-client/src/api-client.ts
fertandil87: I am still using it in const dataset = await GFWAPI.fetch<Dataset>(`/datasets/${LATEST_CARRIER_DATASET_ID}`, { version: '' })

I dont know if it will be needed in another apps

apps/api-portal/features/user-applications/user-applications.ts
```diff
@@ -1,32 +1,24 @@
...
 async function fetchUserApplications(userId: number, limit = 0, offset = 0) {
   const query = stringify({
     'user-id': userId,
     ...((limit && { limit }) || {}),
     ...((offset && { offset }) || {}),
   })
-  const url = `/v2/auth/user-applications?${query}`
-  const data = await GFWAPI.fetch<UserApplicationFetchResponse>(url).catch((error) => {
+  const url = `/auth/user-applications?${query}`
+  const data = await GFWAPI.fetch<APIPagination<UserApplication>>(url).catch((error) => {
```
rdgfuentes, line 21: 👍

PR #1613, opened 2022-08-24
Toggle layers for VV map not showing // VV-441

fertandil87:
wip - is not loading by default the layers from the url


libs/dataviews-client/src/resolve-dataviews-generators.ts
```diff
@@ -545,28 +554,29 @@ export function getDataviewsGeneratorConfigs(
...
+      mergedHeatmapGeneratorId: MERGED_DETECTIONS_ANIMATED_HEATMAP_GENERATOR_ID,
+    })
     : []
-
   const generatorsConfig = [
     ...mergedActivityDataview,
     ...mergedDetectionsDataview,
-    ...otherDataviews,
+    ...trackDataviews,
+    ...otherDataviews
```
rdgfuentes, line 579: Why do we need to separate `trackDataviews` if we don't do any change on them and finally we concat them? They were previously being assigned inside `otherDataviews`

rdgfuentes (approved):

PR #1623, opened 2022-08-25
Added gtag pageview event to tokens site. API-390

rdgfuentes:


rdgfuentes:
Not necessary, fixed using TagManager configuration

PR #1642, opened 2022-08-30
Improve the landing page during vessel searches in vessel viewer // VV-451

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-451

image

### DIFF
--- a/apps/vessel-history/features/search/search.selectors.ts
+++ b/apps/vessel-history/features/search/search.selectors.ts
@@ -5,7 +5,7 @@ import { getSerializedQuery } from './search.thunk'
 
 export const selectVesselsFound = (state: RootState) => state.search.queries
 export const selectSearchError = (state: RootState) => state.search.error
-export const selectSearchSources = (state: RootState) => ['AIS']
+export const selectSearchSources = (state: RootState) => state.search.sources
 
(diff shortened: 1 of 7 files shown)


rdgfuentes (changes requested):
LGTM however I could not find where are we displaying a different message when the vessel search timed out 
> If the search times out, it should say “the search has timed out, please try again

apps/vessel-history/features/search/search.selectors.ts
```diff
@@ -4,6 +4,8 @@ import { selectAdvancedSearchFields, selectQueryParam } from 'routes/routes.sele
 import { getSerializedQuery } from './search.thunk'
 
 export const selectVesselsFound = (state: RootState) => state.search.queries
+export const selectSearchError = (state: RootState) => state.search.error
+export const selectSearchSources = (state: RootState) => ['AIS']
```
rdgfuentes, line 8: Why is this selector returning always `['AIS']`? shouldn't it get the sources from the API response?

fertandil87: ups, I added to test, nice catch kkk

rdgfuentes (approved):
👍

PR #1644, opened 2022-08-31
Display identity and event icons alongside the categories in vessel viewer risk summary tab VV-448

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-448

image


rdgfuentes:
@fertandil87 I think the icons should have the same background that they have in the activity and map tabs for consistency, right?

PR #1663, opened 2022-09-08
Display AIS coverage metric in vessel viewer to a whole number VV-445

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-445

image


rdgfuentes (approved):
🎖️

PR #1675, opened 2022-09-14
Add time range to Risk Summary and Map tab in VV insurance view. VV-492

rdgfuentes:


apps/vessel-history/features/date-range-label/date-range-label.tsx
```diff
@@ -0,0 +1,41 @@
...
+import FiltersLabel from 'features/filters-label/filters-label'
+import { selectCurrentOfflineVessel } from 'features/vessels/offline-vessels.selectors'
+import styles from './date-range-label.module.css'
+
+export interface DateRangeLabelProps {
+  className?: string
+  type?: ButtonType
+}
+
+export function DateRangeLabel({ className, type }: DateRangeLabelProps) {
```
fertandil87, line 18: why this format and not...
const DateRangeLabel: React.FC< DateRangeLabelProps > = (props): React.ReactElement => {

fertandil87 (approved):

apps/vessel-history/features/date-range-label/date-range-label.tsx
rdgfuentes: It's the bolierplate generated by nx for a next component when you run:
```
nx generate @nrwl/next:component DateRangeLabel --project=vessel-history --directory=features --skipTests --no-interactive
```

fertandil87: but... is what I mean is if we should keep always the same syntax

PR #1700, opened 2022-09-23
Setup dataviews for any environment (by slug instead of ids). VV-497

rdgfuentes:


rdgfuentes:
Closed since we're following this other approach #1702

PR #1721, opened 2022-10-03
Add a link in the VV Map tab to view the vessel in the GFW Map VV-421

fertandil87:
still pending some definitions 
refactored code to unify Vessel viewer and fishing map
https://globalfishingwatch.atlassian.net/browse/VV-421


rdgfuentes:
@fertandil87 We reviewed the approach with @j8seangel and we think it would be better to move the logic to generate the vessel url for `fishing-map` to the shared lib `dataviews-client` so that it could be reused from both places (`fishing-map` and `vessel-history`)

rdgfuentes (commented):
Check my comments

apps/vessel-history/features/map/map.hooks.ts
```diff
@@ -53,3 +64,93 @@ export interface Viewport extends LatLon {
...
+    }
+    return null
+  })
+
+  const urlJson: BaseUrlWorkspace = {
+    latitude: viewport.latitude,
+    longitude: viewport.longitude,
+    zoom: viewport.zoom,
+    start: DEFAULT_WORKSPACE.availableStart,
+    end: DEFAULT_WORKSPACE.availableEnd,
```
j8seangel, line 150: Why not use the current user date filter selection and leave the DEFAULT_WORKSPACE as a fallback if the user didn't update the filter?

apps/vessel-history/features/map/map.hooks.ts
```diff
@@ -53,3 +64,93 @@ export interface Viewport extends LatLon {
   longitude: number
   zoom: number
 }
+
+export const openFishingMap = (datasets, allGFWIds, viewport) => {
+  // colors used in the map
+  const colors = ['#f4511f', '#33b679', '#f09300', '#ffea00', '#9ca4ff']
+  const presenceDataviews = [
```
j8seangel, line 71: is this a naming conflict? why not call it contextDataviews?

apps/vessel-history/features/map/map.hooks.ts
```diff
@@ -53,3 +64,93 @@ export interface Viewport extends LatLon {
   longitude: number
   zoom: number
 }
+
+export const openFishingMap = (datasets, allGFWIds, viewport) => {
+  // colors used in the map
+  const colors = ['#f4511f', '#33b679', '#f09300', '#ffea00', '#9ca4ff']
```
j8seangel, line 70: I'd grab this from the ui-components list we are actually using in fishing-map
`LineColorBarOptions`

j8seangel (changes requested):
Apart from the comments, did you test that everything runs properly also in fishing-map?

apps/fishing-map/features/datasets/datasets.utils.ts
```diff
@@ -144,40 +144,6 @@ export const getDatasetsInDataviews = (
   )
 }
 
-export const getRelatedDatasetByType = (
```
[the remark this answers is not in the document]
fertandil87: but I updated every import in the fishing map, is the same

apps/fishing-map/features/dataviews/dataviews.utils.ts
```diff
@@ -35,77 +37,11 @@ export const ENVIRONMENTAL_LAYER_PREFIX = 'environment-'
...
-  datasets: VesselInstanceDatasets
-): DataviewInstance<GeneratorType> => {
-  const vesselDataviewInstance = {
-    id: `${VESSEL_DATAVIEW_INSTANCE_PREFIX}${vessel.id}`,
-    ...vesselDataviewInstanceTemplate(TEMPLATE_VESSEL_DATAVIEW_ID),
-    datasetsConfig: getVesselDataviewInstanceDatasetConfig(vessel.id, datasets),
-  }
-  return vesselDataviewInstance
-}
-
```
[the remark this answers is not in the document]
fertandil87: ok is a nice thing to know, but update all the imports if fine too

apps/vessel-history/features/map/map.hooks.ts
fertandil87: I dont use the current because:
1 is not defined yet
2  will use the voyage dates and is a period so small

apps/vessel-history/features/map/map.hooks.ts
```diff
@@ -53,3 +64,93 @@ export interface Viewport extends LatLon {
   longitude: number
   zoom: number
 }
+
```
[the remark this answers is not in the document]
fertandil87: I dont like to move this function to the shared libs because it has configuration of how leah want to open the link, I am not sure if another app need to use something similar how I will parametrize this. after split the function I will see if any of the functions make sense to move it

PR #1740, opened 2022-10-20
Fixed  Small scale vessel's tracks are being simplified too much that they cross land. MAP-997

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/MAP-997
* Adjust tracks thinning params in km
* Documented thinning config properties


j8seangel (approved):
Great we finally understand what the bearing values are, thanks!

PR #1785, opened 2022-11-18
Upgrade to nx 15 and next 13

rdgfuentes:


apps/fourwings-explorer/features/datasets/DatasetsLibraryLocal.tsx
```diff
@@ -39,7 +38,13 @@ const LocalDatasetsLibrary = ({ datasets }: { datasets: APIDataset[] }) => {
           const disabled = layers.some((l) => l.id === dataset.id)
           return (
             <li key={dataset.id} className={cx(styles.dataset, { [styles.disabled]: disabled })}>
-              {dataset.image && <Image className={styles.image} src={dataset.image}></Image>}
+              {dataset.image && (
+                <img
```
rdgfuentes, line 42: We could continue using the legacy `next/image` component, look there's event a migration tool to make the changes for us https://nextjs.org/blog/next-13#upgrading-nextimage-to-nextjs-13

j8seangel: It not because of the upgrade of the next13. Seems like the Image Loader is [not supported](https://nextjs.org/docs/advanced-features/static-html-export#unsupported-features) by the export static html feature, so there isn't much we can do for now.

PR #1787, opened 2022-11-18
add click tooltip to VV map to show region names VV-477

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-477

image

### DIFF
--- a/apps/vessel-history/features/map/Map.tsx
+++ b/apps/vessel-history/features/map/Map.tsx
@@ -21,9 +21,8 @@ import useViewport from './map-viewport.hooks'
 import MapControls from './controls/MapControls'
 import useMapEvents from './map-events.hooks'
 import { selectHighlightedEvent } from './map.slice'
-import styles from './Map.module.css'
-//import '@globalfishingwatch/maplibre-gl/dist/maplibre-gl.css'
 import PopupWrapper from './popups/PopupWrapper'
+import styles from './Map.module.css'
 
 const mapStyles = {
   width: '100%',
--- a/apps/vessel-history/features/map/Map.tsx
+++ b/apps/vessel-history/features/map/Map.tsx
@@ -22,6 +22,8 @@ import MapControls from './controls/MapControls'
 import useMapEvents from './map-events.hooks'
 import { selectHighlightedEvent } from './map.slice'
 import styles from './Map.module.css'
+import '@globalfishingwatch/maplibre-gl/dist/maplibre-gl.css'
+import PopupWrapper from './popups/PopupWrapper'
 
 const mapStyles = {
   width: '100%',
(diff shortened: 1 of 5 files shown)


apps/vessel-history/features/map/popups/PopupWrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+  latitude: number
+  longitude: number
+}
+function PopupWrapper({ layers, className = '', latitude = 0, longitude = 0 }: PopupWrapperProps) {
+  const getLayerColor = useCallback(
+    (datasetId: string) => {
+      if (datasetId.includes('public-eez')) {
+        return dataviewInstances.find((layerConfig) => layerConfig.id === 'public-eez-areas')
+          ?.config.color
+      }
```
rdgfuentes, line 21: It would be cleaner if we don't have special cases, would be possible to rename the `dataviewInstance.id` to `public-eez-land` so that it matches the `datasetId` to get rid of this special case?

rdgfuentes (dismissed):
Looks good. Only a couple of details to review:
* Clicking on vessel events on the map opens the popup with the context layer info too and it should not.
* Clean up the `public-eez` special case from the PopupWrapper component (check [my comment](https://github.com/GlobalFishingWatch/frontend/pull/1787/files#r1030363364))

apps/vessel-history/features/map/popups/Popup.module.css
```diff
@@ -0,0 +1,143 @@
...
+  min-height: 2rem;
+}
+
+.rowTitle {
+  font: var(--font-XS);
+  text-transform: uppercase;
+  color: var(--color-secondary-blue);
+}
+
+.viirsTable {
```
j8seangel, line 82: there are some unused styles here, could we keep it as small as possible?

apps/vessel-history/features/map/map-events.hooks.ts
```diff
@@ -32,14 +31,19 @@ export default function useMapEvents() {
...
+  const [cursorLayers, setCursorLayers] = useState<ExtendedFeature[] | null>()
 
   const selectVesselEventOnClick = useCallback(
     (event: InteractionEvent | null) => {
       const features = event?.features ?? []
-
       const vesselFeature = features.find(
         (feature) => feature.generatorType === GeneratorType.VesselEvents
       )
+      setCursorLayers(
```
j8seangel, line 42: why not store all the features and render the content based on the generatorType in the [PopupWrapper](https://github.com/GlobalFishingWatch/frontend/blob/develop/apps/fishing-map/features/map/popups/PopupWrapper.tsx#L140) this was we are flexible for different layers in the future

apps/vessel-history/features/map/Map.tsx
```diff
@@ -22,6 +22,8 @@ import MapControls from './controls/MapControls'
 import useMapEvents from './map-events.hooks'
 import { selectHighlightedEvent } from './map.slice'
 import styles from './Map.module.css'
+//import '@globalfishingwatch/maplibre-gl/dist/maplibre-gl.css'
```
j8seangel, line 25: not needed

PR #1805, opened 2022-11-28
Changed rfmo dataset to public-rfmo

rdgfuentes:


libs/i18n-labels/en/datasets.json
```diff
@@ -2506,6 +2506,11 @@
       "vessel_id": "vessel_id"
     }
   },
+  "public-rfmo": {
```
j8seangel, line 2509: did you update this by hand? won't it be overwritten by the automated workflow of the datasets translations

rdgfuentes: you are right, I'll remove them

PR #1811, opened 2022-11-29
Chore: improved per app changelog generation

rdgfuentes:
* Improved changelog generation reading every PR labels and including them in `What's Changed` section only when it belong to the app/lib the changelog is being generated for.


j8seangel:
I tested running `nx changelog fishing-map` and the changelog included PRs like this:

 - [#1803](https://github.com/GlobalFishingWatch/frontend/pull/1803): Download vessel activity in CSV format. VV-514
 
where there no fishing-map label, any idea why?

PR #1825, opened 2022-12-08
dont export fishing-map app

j8seangel:
Testing to skip the export step in builds


rdgfuentes (approved):
@j8seangel LGTM 👍 , could you also apply it to the other apps that have been migrated: `vessel-history`, `api-portal`, `port-labeler`?

PR #1829, opened 2022-12-08
request areas in parallel and fix dataset id in production

j8seangel:
Looks like the `public-rfmo` dataset is not ready for production yet, so this PR uses the legacy dataset until we have it ready.
Also took advantage of the change to make this faster requesting the 3 areas in parallel, please @rdgfuentes take a look at this later as I'm going to merge this to get the bug fixed asap.


rdgfuentes (commented):
LGTM 👍

PR #1852, opened 2022-12-16
Added new analytics events to the map

fertandil87:


apps/fishing-map/features/workspace/activity/ActivityFilters.tsx
```diff
@@ -122,6 +122,7 @@ function ActivityFilters({ dataview: baseDataview }: ActivityFiltersProps): Reac
 
   const onDataviewFilterChange = useCallback(
     (dataviewInstance: UrlDataviewInstance) => {
+      console.log(newDataviewInstanceConfig)
```
j8seangel, line 125: 👀

libs/timebar/src/timebar.js
```diff
@@ -292,6 +292,7 @@ class Timebar extends Component {
               absoluteStart={absoluteStart}
               absoluteEnd={absoluteEnd}
               onBookmarkChange={this.props.onBookmarkChange}
+              onIntervalClick={this.props.onIntervalClick}
```
j8seangel, line 295: This function says in the label:
```
category: 'Timebar'
action: 'Bookmark timerange'
label: 'removed'
```
could you update it?

apps/fishing-map/features/workspaces-list/WorkspaceWizard.tsx
```diff
@@ -85,9 +88,15 @@ function WorkspaceWizard() {
     } else {
       updateMatchingAreas(inputValue)
     }
+    setInputSearch(inputValue)
```
j8seangel, line 91: why do we need to replicate this? we already have the input value in [this param](https://github.com/GlobalFishingWatch/frontend/blob/add-analytics-mm/apps/fishing-map/features/workspaces-list/WorkspaceWizard.tsx#L143): `inputValue`
You could move the function below the hook to use it

j8seangel (changes requested):

apps/fishing-map/features/workspaces-list/WorkspaceWizard.tsx
fertandil87: oh this PR wasn't done, but in this case.. the reason is because the inputValue has the full name of the selected object at the moment that an item is selected form the list, and I want to capture what the user type and what the user select @j8seangel

apps/fishing-map/features/workspace/activity/ActivityFilters.tsx
fertandil87: yes yes, I didn't add you as reviewer, it wasn't done

libs/timebar/src/timebar.js
fertandil87: this line is not needed, removed

apps/fishing-map/features/workspaces-list/WorkspaceWizard.tsx
j8seangel: got it, thanks!

j8seangel (approved):

PR #1858, opened 2023-01-09
Api authentication with next. VV-520

rdgfuentes:


apps/fishing-map/store.ts
```diff
@@ -89,10 +89,10 @@ const store = configureStore({
   reducer: rootReducer,
   middleware: (getDefaultMiddleware) =>
     getDefaultMiddleware(defaultMiddlewareOptions)
+      .concat(dataviewStatsApi.middleware)
```
j8seangel, line 92: 🙇

libs/authjs-client/src/lib/gfw-client.ts
```diff
@@ -0,0 +1,365 @@
...
+type RequestData = {
+  requestData: {
+    url: string
+    method: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH'
+    body?: string
+  }
+  headers: Headers
+}
+const defaults: GFWConfig = {
+  gatewayUrl: 'https://gateway.api.globalfishingwatch.org',
```
j8seangel, line 28: I'd add the env variable and use this as default in the same way we do [here](https://github.com/GlobalFishingWatch/frontend/blob/feature/VV-520-api-authentication-with-next/libs/api-client/src/api-client.ts#L12)

libs/authjs-client/src/lib/gfw-client.ts
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  /**
+   * Send a PATCH request
+   * @param resource - endpoint
+   * @param parameters optional parameters to be send in the query
+   * @param body - PUT parameters object.
+   * @returns Promise resolving to the response from the API.
+   */
+  patch(
```
j8seangel, line 345: I like this!

j8seangel (commented):
Edit: I missed the `NEXTAUTH_SECRET` env var and it works now

Looking great but I couldn't manage to get it working in fishing-map app as I got error 500 in every API auth request. Am I missing something? This is the log error: 
```
TypeError: Converting circular structure to JSON
```
I did a commit with some of the changes and merged with develop to get it working again after the terraform changes

j8seangel:
Closing as we can't use auth in the server yet

PR #1864, opened 2023-01-12
Deck playground/recursive layer sorting

weberjavi:
This PR creates a sort function to order layers before passing them to Deck


apps/deck-playground/layers/fourwings/FourwingsHeatmapTileLayer.ts
```diff
@@ -29,6 +29,7 @@ export type FourwingsHeatmapTileLayerProps = {
   resolution?: FourwingsLayerResolution
   minFrame: number
   maxFrame: number
+  zOrderIndex: number
```
j8seangel, line 32: What about making this prop optional? so we can use by default the `GROUP_ORDER` defined in the const but allow to override if needed.

apps/deck-playground/layers/fourwings/FourwingsLayer.ts
```diff
@@ -41,7 +41,7 @@ export class FourwingsLayer extends CompositeLayer<FourwingsLayerProps & TileLay
   }
 
   getMode() {
-    return this.props.mode || 'heatmap'
+    return this.props.mode || HEATMAP_ID
```
j8seangel, line 44: 👏 thanks!

apps/deck-playground/layers/fourwings/FourwingsPositionsTileLayer.ts
```diff
@@ -22,6 +22,7 @@ import { FourwingsColorRamp } from './FourwingsLayer'
 export type FourwingsPositionsTileLayerProps<DataT = any> = {
   minFrame: number
   maxFrame: number
+  zOrderIndex: number
```
j8seangel, line 25: why not just zIndex ? does the order add something?

apps/deck-playground/layers/vessel/VesselsLayer.ts
```diff
@@ -4,7 +4,7 @@ import { VesselLayer, VesselLayerProps } from 'layers/vessel/VesselLayer'
 export type VesselsLayerProps = { ids: string[] } & VesselLayerProps
 
 export class VesselsLayer extends CompositeLayer<VesselsLayerProps> {
-  vesselLayers = this.props.ids.map(
+  layers = this.props.ids.map(
```
j8seangel, line 7: As this needs to be a convention, where do you think we could add the documentation?

j8seangel (commented):
I love the solution you found! works great and allow us to keep using classes for cleaner code structure.

weberjavi:
@j8seangel thanks for your comments. Agreed on all of them 😃 
I've added README with some context on the intention of the app, scope and some conventions. In the future we might also want to add some content regarding layers architecture.

Let me know if you see something that could be improved 🙇🏼

j8seangel:
So nice we have a README as a starting point, it looks great!

The only thing may worth discussion is to create a generic custom composite layer type definition to enforce we return layers as an array of sublayers, so we could do something like:

```
export class VesselLayer extends GFWCompositeLayer
```

where `GFWCompositeLayer` just include layers[] as a mandatory property

WDYT ?

PR #1874, opened 2023-01-20
Feature/api auth with next review

j8seangel:
To avoid requesting user data `/me` twice this PR removes the fetch from the provider and do in the server side using `getServerSideProps` 

Also took advantage to clean and organise a bit the authjs-client code separating in different files like:
- config.ts
- types.ts
- utils.ts


libs/authjs-client/src/lib/gfw-provider.ts
```diff
@@ -59,26 +57,28 @@ export function GFWProvider<P extends GFWProfile>(
...
       return {
-        id: `${profile.id}`,
-        name: profile.firstName + ' ' + profile.lastName,
-        email: profile.email,
-        image: profile.photo,
+        id: profile.data.id,
+        name: profile.data.firstName + ' ' + profile.data.lastName,
+        email: profile.data.email ?? '',
+        image: profile.data.photo ?? '',
       }
```
rdgfuentes, line 66: 👍

libs/authjs-client/src/lib/gfw-client.ts
```diff
@@ -1,119 +1,8 @@
...
-    method: requestData.method,
-    headers: Object.assign({}, baseHeaders, headers) as any,
-  }).then(_handleResponseTextOrJson)
-
-  return results
-}
+import type { UserData, UserPermission, APIPagination } from '@globalfishingwatch/api-types'
+import { GFW_API_AUTH_ENDPOINT, GFW_CLIENT_BASE_HEADERS, GFW_CLIENT_DEFAULTS } from './config'
+import { GFWConfig, GFWConfigParams, RequestData, RequestDataMethod, Token } from './types'
+import { handleResponseTextOrJson } from './utils'
```
rdgfuentes, line 4: Thanks for refactoring this into a separate file 👍

libs/authjs-client/src/lib/gfw-client.ts
```diff
@@ -136,74 +21,119 @@ export class GFW {
...
+      })
+      const user: UserData = { id: 0, type: GUEST_USER_TYPE, permissions, groups: [] }
+
+      return user
+    } catch (e) {
+      if (this.config.debug) {
+        console.log('GFW-API. Error fetching guest user', e)
+      }
+      throw e
+    }
```
rdgfuentes, line 136: I was reluctant to add these methods in the client because I didn't want to couple it to the `@globalfishingwatch/api-types` lib. However I understand that they will be for sure reused in other apps consuming this client. So I think we could leave it here unless you find a better place for them. WDYT?

j8seangel: As we need the `fetchGuestUser` method to return the user in this format:
```js
{ id: 0, type: GUEST_USER_TYPE, permissions, groups: [] }
```
The only way I think we could remove the dependency from the `api-types` is adding a generic to the fetchGuestUser<User = unknown> and consume like:
```js
client.fetchGuestUser<UserData>()
```
where needed, for example  [here](https://github.com/GlobalFishingWatch/frontend/blob/feature/api-auth-with-next-review/apps/fishing-map/pages/index.tsx#L28) WDYT ?

rdgfuentes:
@j8seangel Great improvements! It's great to start taking advantage of server side capabilities.

PR #1875, opened 2023-01-24
Deck playground/context layers

weberjavi:
This PR updates deckgl version to `8.8.24` to avoid the [antimeridian bug](https://github.com/visgl/deck.gl/issues/7614) and also adds `jotai` as a dependency, so please do `yarn` before testing it.

We are adding a couple of layers here: 
- **ContextLayer**, it is divided in three different layers to allow painting the lines and fills of the highlighted features in the z index they belong to.
- **Basemap**, the layer has been split into bathimetry and landmass to adecuate to the needed z ordering 

The main update is that we are handling events through the `onHover` and `onClick` callbacks of the `deck` instance in order to be able to `pickMultipleObjects`. With this approach we should avoid using the `getPickingInfo` callback of the layers in order to avoid performing picking twice (read [this discussion](https://github.com/visgl/deck.gl/discussions/5793) to get more context ).


satellitestudiodesign (commented):
Looks and performs great 👏

Some minor stuff (we can address it now or just take note for when we incorporate the layers in the map):
- Looks like both the lines and the fills are pickable and you see both on the tooltip when the cursor is exactly over a line.
- I'm getting a lot of `Warning: Encountered two children with the same key...` from InfoTooltip (looks related with the one above, maybe fills and lines have the same key)
- The z-order of the lines when they're not highlighted should be below the landmass

j8seangel:
I've been playing with different debounced times and I think we could benefit a lot in performance when needed but I think this is good to go for now! Great work @weberjavi 🙇

PR #1877, opened 2023-01-30
update yarn version ENG-444

j8seangel:
Testing https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern

Blockers found:
- [x] Looks like tslib is not being imported properly as building apps returns this error:
```
This syntax requires an imported helper named '__spreadArray' which does not exist in 'tslib'. Consider upgrading your version of 'tslib'
```
- [x] The same with some types, for example in fishing-map:
```
 ./pages/index.tsx:18:6
       Type error: Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes'.
```
- [x] Running scripts between apps is also broken:
```
Usage Error: The nearest package directory (/Users/joseangel/Workspace/frontend-yarn/libs/i18n-labels) doesn't seem to be part of the project declared in /Users/joseangel/Workspace/frontend-yarn.
- If /Users/joseangel/Workspace/frontend-yarn isn't intended to be a project, remove any yarn.lock and/or package.json file there.
- If /Users/joseangel/Workspace/frontend-yarn is intended to be a project, it might be that you forgot to list libs/i18n-labels in its workspace configuration.
- Finally, if /Users/joseangel/Workspace/frontend-yarn is fine and you intend libs/i18n-labels to be treated as a completely separate project (not even a workspace), create an empty yarn.lock file in it.
```

One of the main advantages:
image


package.json
```diff
@@ -29,8 +29,7 @@
     "clean:deps": "rm -rf node_modules/.yarn-integrity && yarn install",
     "clean:dist": "rm -rf dist",
     "clean": "yarn clean:deps && yarn clean:dist",
-    "prepare": "husky install",
-    "postinstall": "yarn yarn-deduplicate yarn.lock"
```
fertandil87, line 33: when I was testing I needed not delete, but update to:

postinstall: "yarn dedupe"

fertandil87: did you see this @j8seangel ?

j8seangel: @fertandil87 yarn 3 is doing this natively so we don't need to run the script anymore

rdgfuentes (approved):
LGTM

PR #1881, opened 2023-01-31
fix fishing-map errors ENG-444

fertandil87:


apps/port-labeler/package.json
```diff
@@ -33,9 +33,9 @@
     "react-virtualized-auto-sizer": "^1.0.5",
     "react-window": "^1.8.6",
     "recoil": "0.7.5",
-    "redux": "4.2.0",
-    "redux-first-router": "2.1.5",
-    "redux-first-router-link": "2.1.1",
+    "redux": "^4.1.2",
```
rdgfuentes, line 36: @fertandil87 Why are you using an older version of redux?

package.json
```diff
@@ -8,7 +8,8 @@
   "workspaces": {
     "packages": [
       "apps/*",
-      "linting"
+      "linting",
+      "libs/i18n-labels"
```
rdgfuentes, line 12: @fertandil87 Why do we need to add this?

package.json
```diff
@@ -32,6 +33,7 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@globalfishingwatch/i18n-labels": "*",
```
rdgfuentes, line 36: @fertandil87 Do we need this? why?

j8seangel:
@fertandil87 I still had the same issue but updating all the dependencies has fixed it. Merging with the base branch to test everything there before merging

PR #1896, opened 2023-02-10
organize dependencies ENG-444

fertandil87:
unify common dependencies or duplicated in all projects in the main package


package.json
```diff
@@ -65,6 +66,7 @@
     "gl-matrix": "3.3.0",
     "lodash": "4.17.21",
     "luxon": "^2.5.2",
+    "mapbox-gl": "^2.12.1",
```
rdgfuentes, line 69: Why adding this dependency? We are using a fork of mapbox-gl. In other places we are adding an empty dependency if it's needed:
```
"mapbox-gl": "npm:empty-npm-package@1.0.0",
```

fertandil87: because is a dependency of react-map-gl and to avoid a warning, I can remove it

PR #1897, opened 2023-02-13
Dynamic report tasks 

satellitestudiodesign:


apps/fishing-map/features/reports/reports.selectors.ts
```diff
@@ -95,19 +96,32 @@ export const selectReportVesselsGraphData = createSelector(
...
+        if (b.name === DEFAULT_NULL_VALUE) return -1
+        return sum(dataviewIds.map((d) => b[d])) - sum(dataviewIds.map((d) => a[d]))
+      })
+
+    if (distributionKeys.length <= MAX_CATEGORIES) return data
+
+    const top = data.slice(0, MAX_CATEGORIES)
+    const rest = data.slice(MAX_CATEGORIES)
+    const others = {
+      name: 'others',
```
j8seangel, line 119: I think we'll need to translate this

PR #1898, opened 2023-02-14
use new refreshToken after updating

j8seangel:
Update api-client to match [new refreshToken policy](https://globalfishingwatch.atlassian.net/browse/API-461)


rdgfuentes:
👍

PR #1901, opened 2023-02-16
Editable layer

weberjavi:


apps/deck-playground/layers/custom-reference/CustomReferenceLayer.ts
```diff
@@ -0,0 +1,19 @@
...
+  renderLayers() {
+    const { mode, data, selectedFeatureIndexes, onEdit, onClick } = this.props
+    return [
+      new EditableGeoJsonLayer({
+        id: 'custom-reference',
+        data,
+        onEdit,
+        onClick,
+        selectedFeatureIndexes,
+        mode: mode === 'edit' ? ModifyMode : DrawPolygonMode,
```
j8seangel, line 15: I got this error `Deprecated use of passing `mode` as a string. Pass the mode's class constructor instead.` but as I see here you are not using a string. any idea what is going on?

apps/deck-playground/layers/custom-reference/CustomReferenceLayer.ts
```diff
@@ -0,0 +1,19 @@
+import { EditableGeoJsonLayer } from '@nebula.gl/layers'
+import { DrawPolygonMode, ModifyMode } from '@nebula.gl/edit-modes'
+
+export class CustomReferenceLayer extends EditableGeoJsonLayer {
+  static layerName = 'custom-reference-layer'
+  renderLayers() {
+    const { mode, data, selectedFeatureIndexes, onEdit, onClick } = this.props
+    return [
+      new EditableGeoJsonLayer({
```
j8seangel, line 9: is there an option to drag the polygon?

j8seangel (approved):
It looks and works great!! 🥇 

We are missing the feature of fine-tunning a point coordinate but as I saw we have the onClick callback and the data available in the atom will be easier to include.

And not directly related to this code but maybe to the deck.gl version. any idea why this warning appears in the console now?
```deck: FourwingsHeatmapLayer.componentName not specified```

apps/deck-playground/layers/custom-reference/CustomReferenceLayer.ts
weberjavi: Looking into this 👁️ 🐽 👁️

PR #1910, opened 2023-02-22
change page description for marine manager

j8seangel:
Render different page description for Marine Manager workspaces

### DIFF
--- a/apps/fishing-map/pages/index.tsx
+++ b/apps/fishing-map/pages/index.tsx
@@ -1,6 +1,8 @@
 import path from 'path'
+import Head from 'next/head'
 import dynamic from 'next/dynamic'
 import { MapProvider } from 'react-map-gl'
+import { Component, Fragment } from 'react'
 
 // This is needed by nx/next builder to run build the standalone next app properly
 // https://github.com/nrwl/nx/issues/9017#issuecomment-1140066503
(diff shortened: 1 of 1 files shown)


apps/fishing-map/pages/index.tsx
```diff
@@ -14,11 +16,51 @@ const AppNoSSRComponent = dynamic(() => import('../features/app/App'), {
...
 
-const Index = () => {
+const HeadDescriptionByLocation = ({ isMarineManagerLocation }) => {
   return (
-    <MapProvider>
-      <AppNoSSRComponent />
-    </MapProvider>
+    <Head>
+      <meta
+        property="og:description"
```
rdgfuentes, line 23: There are already meta tags on the _document.tsx file https://github.com/GlobalFishingWatch/frontend/blob/22a402c5393346f456e01e3e81022880c2d77a7b/apps/fishing-map/pages/_document.tsx#L16

j8seangel: Based on [this documentation](https://nextjs.org/docs/api-reference/next/head) `In this case only the second <meta property="og:title" /> is rendered. meta tags with duplicate key attributes are automatically handled.` I don't think that will be a problem.
And we need to handle it at index.tsx page as document can't read the originalUrl

PR #1968, opened 2023-03-23
Make a Blueprint version of the Vessel Viewer // VV-550

fertandil87:
https://globalfishingwatch.atlassian.net/browse/VV-550

image

### DIFF
--- a/apps/vessel-history/features/datasets/datasets.slice.ts
+++ b/apps/vessel-history/features/datasets/datasets.slice.ts
@@ -1,9 +1,9 @@
 import { createAsyncThunk, createSelector } from '@reduxjs/toolkit'
 import { uniqBy, memoize, without } from 'lodash'
 import { stringify } from 'qs'
+import { GFWApiClient } from 'http-client/http-client'
 import { APIPagination, Dataset, DatasetTypes, RelatedDataset } from '@globalfishingwatch/api-types'
 import {
-  GFWAPI,
   parseAPIError,
   parseAPIErrorMessage,
   parseAPIErrorStatus,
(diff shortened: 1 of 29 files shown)


apps/vessel-history/features/datasets/datasets.slice.ts
```diff
@@ -136,16 +140,20 @@ export const selectAll = createSelector([baseSelectAll], (datasets) => {
...
+            ...d,
+            relatedDatasets: [
+              ...d.relatedDatasets,
+              {
+                id: 'proto-global-gaps-events:v20201001',
+                type: DatasetTypes.Events,
+              },
+            ],
+          }
+    )
```
rdgfuentes, line 156: It would be more convenient to verify this only once and not on every dataset mapping since the logic will not change for different datasets. I'd extract this callback to a function and use it here, eg: 
```typescript
  const mapDataset =  IS_STANDALONE_APP
        ? (d) => d
        : (d) => {
            ...d,
            relatedDatasets: [
              ...d.relatedDatasets,
              {
                id: 'proto-global-gaps-events:v20201001',
                type: DatasetTypes.Events,
              },
            ],
          }
  const vesselInfo: Dataset[] = datasets
    .filter((d) => d.category === 'vessel' && d.subcategory === 'info')
    // Inject Proto Gaps Dataset
    .map(mapDataset)
```

fertandil87: this code is not needed, it was a path to support gaps, now is included in the dataview

apps/vessel-history/features/dataviews/dataviews.slice.ts
```diff
@@ -18,12 +19,27 @@ export const fetchDataviewsByIdsThunk = createAsyncThunk(
...
+            !IS_STANDALONE_APP
+              ? dataview
+              : {
+                  ...dataview,
+                  datasetsConfig: dataview.datasetsConfig?.filter(
+                    (conf) =>
+                      conf.datasetId.startsWith('public-') && conf.endpoint !== DatasetTypes.Tracks
+                  ),
+                }
+          )
```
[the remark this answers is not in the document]
fertandil87: with the new dataview this is not needed

rdgfuentes (approved):
LGTM

PR #1982, opened 2023-03-29
add user dataset to current workspace in user panel

j8seangel:
Adds a new arrow icon to see the selected dataset in the current workspace. (If there isn't one, it will use the default one)

image


satellitestudiodesign (approved):
cool cool cool

PR #2012, opened 2023-04-21
Support workspaces with less delay than default of 3 days

j8seangel:
Now the app supports a workspace with an `endAt` bigger than the `LAST_DATA_UPDATE` to fix the `private_papua_new_guinea` workspace

image


j8seangel:
As endAt needs to be dynamic it will be addressed in a new PR

PR #2013, opened 2023-04-21
Fishing map e2e/test login

j8seangel:
WIP: testing login with cypress 

image

Captura de Pantalla 2023-07-21 a la(s) 20 18 12


j8seangel (commented):
@rdgfuentes such a great work caching the dependencies and integrating in the workflow 🙇

nataliafz (approved):

fertandil87 (approved):

apps/fishing-map-e2e/cypress.config.ts
```diff
@@ -8,7 +8,7 @@ const cypressJsonConfig = {
   videosFolder: '../../dist/cypress/apps/fishing-map-e2e/videos',
   screenshotsFolder: '../../dist/cypress/apps/fishing-map-e2e/screenshots',
   chromeWebSecurity: false,
-  specPattern: 'src/e2e/**/*.cy.{js,jsx,ts,tsx}',
+  specPattern: ['src/e2e/app.cy.ts', 'src/e2e/login.cy.ts', 'src/e2e/**/*.cy.{js,jsx,ts,tsx}'],
```
j8seangel, line 11: @nataliafz do we need to define every test case file here?

nataliafz: @j8seangel no, with 'src/e2e/**/*.cy.{js,jsx,ts,tsx}' should be enough

j8seangel: I'd remove the fixed ones then to avoid confusion

fertandil87: without that won't work @j8seangel @rdgfuentes, the reason is that in login we are removing a popup and then in vms we dont need to remove it again because the localstorage is not shown again, if the e2e modify this order the popup appear and the test fail. 
For this I implemented a function in another test that hide all the popups so then I just need to include that in the vms test.
But also my idea is to avoid tests when one fail, so the idea is just to test if the app open, and maybe other basic tests, before the more complex tests that can take more time

apps/fishing-map-e2e/src/e2e/map/map.cy.ts
```diff
@@ -0,0 +1,49 @@
...
+  afterTestSkip,
+  beforeTestSkip,
+  disablePopups,
+  getTimeline,
+  waitForMapLoadTiles,
+} from '../../support/app.po'
+
+describe('map', () => {
+  before(() => beforeTestSkip())
+  afterEach(afterTestSkip)
```
rdgfuentes, line 12: @fertandil87 Why are we skipping before and after tests calls?

apps/fishing-map-e2e/src/e2e/map/map.cy.ts
```diff
@@ -0,0 +1,49 @@
...
+    cy.getByClass('VesselsTable_vesselsTable')
+      .find('tbody')
+      .find('tr')
+      .first()
+      .find('td')
+      .eq(1)
+      .then((vessel) => {
+        const textValue = vessel.text()
+        cy.wrap(textValue).as('firstVessel')
+      })
```
rdgfuentes, line 32: Using class, id or tag selectors as well as navigating the DOM structure in our e2e tests is an [anti-pattern](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements). They are subject to change and will break our tests sooner than later.

apps/fishing-map-e2e/src/e2e/map/map.cy.ts
fertandil87: @rdgfuentes It was a proposal and it was merged... This was already discussed and exist another PR to fix it https://github.com/GlobalFishingWatch/frontend/pull/2147

apps/fishing-map-e2e/cypress.config.ts
```diff
@@ -23,4 +24,9 @@ export default defineConfig({
      **/
     testIsolation: false,
   },
+  env: {
+    apiAuthUser: '',
+    apiAuthPass: '',
+    FAIL_FAST_ENABLED: true,
```
rdgfuentes, line 30: :broom:

apps/fishing-map-e2e/src/e2e/fishing-activity/vms.cy.ts
```diff
@@ -0,0 +1,20 @@
+import { VMS_NAMES } from '../../constants/vms'
+import { disablePopups, getVmsActivityLayerPanel, switchLanguage } from '../../support/app.po'
+
+describe('vms', () => {
+  beforeEach(() => cy.visit('/'))
+  disablePopups()
```
rdgfuentes, line 6: Move inside beforeEach

rdgfuentes (approved):

PR #2014, opened 2023-04-21
Fishing map/latest data available from datasets. MAP-1160

j8seangel:
Fixes https://github.com/GlobalFishingWatch/frontend/pull/2012 using endDate of datasets active in the workspace

### DIFF
--- a/apps/fishing-map/data/config.ts
+++ b/apps/fishing-map/data/config.ts
@@ -51,12 +51,10 @@ export const FULL_SUFIX = 'full'
 export const USER_SUFIX = 'user'
 export const PRIVATE_SUFIX = 'private'
 
+export const DEFAULT_DATA_DELAY_DAYS = 3
 // used when no url data and no workspace data
-export const LAST_DATA_UPDATE = DateTime.fromObject(
-  { hour: 0, minute: 0, second: 0 },
-  { zone: 'utc' }
-)
-  .minus({ days: 3 })
+const LAST_DATA_UPDATE = DateTime.fromObject({ hour: 0, minute: 0, second: 0 }, { zone: 'utc' })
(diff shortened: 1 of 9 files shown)


apps/fishing-map/features/datasets/datasets.utils.ts
```diff
@@ -247,6 +247,20 @@ export const getActiveDatasetsInActivityDataviews = (
   })
 }
 
+export const getLatestEndDateFromDatasets = (
+  datasets: Dataset[],
+  datasetCategory?: DatasetCategory
+): string => {
+  if (!datasets.length) return ''
```
rdgfuentes, line 254: What if there's no visible datasets by default? It'll return `''`, is that expected? Shouldn't it return `DEFAULT_TIME_RANGE.end` in this case to ensure it returns a proper date?

apps/fishing-map/data/config.ts
```diff
@@ -51,12 +51,10 @@ export const FULL_SUFIX = 'full'
 export const USER_SUFIX = 'user'
 export const PRIVATE_SUFIX = 'private'
 
+export const DEFAULT_DATA_DELAY_DAYS = 3
```
rdgfuentes, line 54: 👍

apps/fishing-map/features/app/app.selectors.ts
```diff
@@ -67,6 +72,24 @@ export const selectTimeRange = createSelector(
   }
 )
 
+export const selectLatestAvailableDataDate = createSelector(
```
rdgfuentes, line 75: 💯

apps/fishing-map/features/datasets/datasets.utils.ts
j8seangel: Good catch! thanks

PR #2048, opened 2023-05-09
Default map should have all public vms sources available. e2e. MAP-1169

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/MAP-1169
* Added tests to ensure all public vms sources are available in the apparent fishing effort layer
* Use `data-test` attribute to get select nodes for e2e testing


apps/fishing-map-e2e/src/e2e/fishing-activity/vms.cy.ts
```diff
@@ -0,0 +1,23 @@
...
+
+describe('vms', () => {
+  beforeEach(() => cy.visit('/'))
+
+  it('should include all sources available publicly', () => {
+    // Open the filter popup
+    getVmsActivityLayerPanel().find(`[data-test*="activity-layer-panel-btn-filter-vms-"]`).click()
+
+    // Check that all vms layers are available
+    cy.getBySelLike('activity-filters-option').contains('AIS')
```
j8seangel, line 11: I don't know if this is a good practice, but should we use [this list](https://github.com/GlobalFishingWatch/frontend/blob/master/apps/fishing-map/data/default-workspaces/workspace.production.ts#L64) of datasets instead of maintaining a fixed list here?

PR #2062, opened 2023-05-12
add admin group in list for admins

j8seangel:
This allows to remove yourself as an admin while you are an admin


satellitestudiodesign (approved):
👌

PR #2081, opened 2023-05-31
Load related datasets recursively. VMS-51

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VMS-51

### DIFF
--- a/apps/fishing-map/features/datasets/datasets.slice.ts
+++ b/apps/fishing-map/features/datasets/datasets.slice.ts
@@ -76,7 +76,8 @@ export const fetchDatasetByIdThunk = createAsyncThunk<
 const fetchDatasetsFromApi = async (
   ids: string[] = [],
   existingIds: string[] = [],
-  signal: AbortSignal
+  signal: AbortSignal,
+  maxDepth: number = 5
 ) => {
   const uniqIds = ids?.length ? ids.filter((id) => !existingIds.includes(id)) : []
   const datasetsParams = {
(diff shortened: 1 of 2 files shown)


apps/fishing-map/features/datasets/datasets.slice.ts
```diff
@@ -73,43 +73,56 @@ export const fetchDatasetByIdThunk = createAsyncThunk<
   }
 })
 
+const fetchDatasetsFromApi = async (
+  ids: string[] = [],
+  existingIds: string[] = [],
+  signal: AbortSignal,
+  maxDepth: number = 5
```
j8seangel, line 80: I'd set up this as `maxDepth=2` to ensure we won't go deeper than two levels. 
This way we only go to the related of a related instead of fetching more datasets than probably needed

j8seangel (approved):
There was [something similar](https://github.com/GlobalFishingWatch/frontend/blob/bugfix/VMS-50-load-related-datasets/apps/fishing-map/features/map/map.slice.ts#L267) but just for the infoDatasets.
But I like most the idea of doing it just once at the beginning, LGTM!

apps/fishing-map/features/datasets/datasets.slice.ts
j8seangel: Activity => vessel info => vessel track is

j8seangel: Sounds good, let's keep 5 although I don't think we won't ever need it as the deeper relationship we have is:
`activity => info => track` and I don't think that will change we will be safe in the future

PR #2085, opened 2023-06-05
Deck migration/fourwings initial structure

weberjavi:
## General
- [ ] ~Change mode depending on zoom level~ (some thinking is needed on how user would interact with this)
- [x] Use sidebar toogle to hide/show layer
## Heatmap
- [x] Allow all time intervals
- [x] Update generator 
- [x] Display data on timebar
- [x]  Display data on timebar tooltip
- [ ] Display labels on sidebar
- [ ] Display colorRange on sidebar
- [ ] Combine layers
- [ ] On average aggregation no zero values should be passed to the stackedActivity chart path

## Bugs
- [x] Fix visibility layer toggle behaviour
- [x] timebar behaviour on different timeranges (`NaN` on generated paths)
- [ ] First elements on timebar sometimes do not have al datasets data 👇🏼 

image
image
image


j8seangel (commented):
Let's talk whenever works best for you to have a quick chat to understand the sublayers categorization better.
I also struggle with performance in the timebar hover. We can brainstorm some ideas in the same session

PR #2103, opened 2023-06-22
save data status by type

j8seangel:
As we need granularity for the status of the data vessel resources this PRs replaces the atom `layersLoaded` property by a new `dataStatus` one. 
This stores an array of `{ type: 'track' | EventType , status: ResourceStatus }` with every layer loaded. [See example of how to get the status](https://github.com/GlobalFishingWatch/frontend/compare/deck-migration/layers-hooks-review...deck-migration/layers-hooks-data-status?expand=1#diff-b6ec3c581e2855592974c9e919dea4bdd64e083a4913c63492a8e13afda7518bR84)

Pending TODO:
- [ ] Handle data loading errors as I couldn't catch the loader error
- [ ] Reset status if data URL changes

### DIFF
--- a/apps/fishing-map/features/map/map-layers.hooks.ts
+++ b/apps/fishing-map/features/map/map-layers.hooks.ts
@@ -1,7 +1,6 @@
 import { useSelector } from 'react-redux'
 import { useMemo } from 'react'
-import { useDeckLayerComposer, useVesselLayers } from '@globalfishingwatch/deck-layers'
-import { UrlDataviewInstance } from '@globalfishingwatch/dataviews-client'
+import { useDeckLayerComposer } from '@globalfishingwatch/deck-layers'
 import { useGeneratorsConnect, useGeneratorsDictionaryConnect } from 'features/map/map.hooks'
 import { selectHighlightedTime } from 'features/timebar/timebar.slice'
 import { useHighlightedEventsConnect } from 'features/timebar/timebar.hooks'
@@ -25,11 +24,4 @@ export const useMapDeckLayers = () => {
   })
 
   return layers
-}
-
-export const useMapVesselLayer = (dataview: UrlDataviewInstance) => {
-  const vesselLayers = useVesselLayers()
-  return useMemo(() => {
-    return vesselLayers.find((d) => d.id === dataview.id)
-  }, [dataview, vesselLayers])
 }
--- a/apps/fishing-map/features/map/map-layers.hooks.ts
+++ b/apps/fishing-map/features/map/map-layers.hooks.ts
@@ -1,6 +1,7 @@
 import { useSelector } from 'react-redux'
 import { useMemo } from 'react'
-import { useDeckLayerComposer } from '@globalfishingwatch/deck-layers'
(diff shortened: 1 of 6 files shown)


satellitestudiodesign (approved):

libs/deck-layers/src/layers/vessel/VesselLayer.ts
```diff
@@ -1,108 +1,158 @@
...
+} from '@globalfishingwatch/api-types'
 import { parquetLoader } from '../../loaders/vessels/trackLoader'
 import { VesselDeckLayersEventData, vesselEventsLoader } from '../../loaders/vessels/eventsLoader'
 import { VesselDeckLayersEvent } from '../../layer-composer/types/vessel'
 import { deckToHexColor } from '../../utils/colors'
 import { EVENTS_COLORS, VesselEventsLayer, _VesselEventsLayerProps } from './VesselEventsLayer'
 import { VesselTrackLayer, _VesselTrackLayerProps } from './VesselTrackLayer'
 
+export const TRACK_LAYER_TYPE = 'track'
+export type VesselDataStatus = {
```
weberjavi, line 26: 💯 Nice way of aligning layers data state with general resources management on the platform

apps/fishing-map/features/map/map-layers.hooks.ts
```diff
@@ -25,3 +26,10 @@ export const useMapDeckLayers = () => {
 
   return layers
 }
+
+export const useMapVesselLayer = (dataview: UrlDataviewInstance) => {
```
weberjavi, line 30: 🤔 my first thought is that this hook might belong to layer hooks and not map hooks. 
It is true that consuming a `UrlDataviewInstance` is coupling it with the app, but maybe it can consume an `id:string` and be used from the outside as `useMapVesselLayer(dataview.id)`, that way it might be decoupled from the workspace-dataview-dataviewinstance flow and be reused in other apps that do not rely on that? If this makes any sense 😅

weberjavi (commented):
Added some small comments ✍🏼

apps/fishing-map/features/map/map-layers.hooks.ts
j8seangel: I like the idea, implemented [here](https://github.com/GlobalFishingWatch/frontend/commit/13cbf713005433d8d538dd1ee3b108f475b02094)

libs/deck-layers/src/layers/vessel/VesselLayer.ts
```diff
@@ -1,108 +1,158 @@
...
     return this.layers
   }
 
+  filterSubLayer(context: FilterContext) {
+    const subLayerType = context.layer.props.type
+    if (subLayerType === TRACK_LAYER_TYPE) {
+      return this.props.visible
+    }
+    if (this.props.visibleEvents?.length) {
+      return this.props.visibleEvents.includes(subLayerType)
```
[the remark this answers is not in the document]
j8seangel: I see the advantage of requesting data conditionally based on layer visibility, so we don't load data we might not be used. Will try to implement, but anyway I think using this flag should be more performant as the shaders shouldn't be in use.

PR #2116, opened 2023-06-29
real-time-prototype tasks

satellitestudiodesign:
- [ ] Log in again when a fetch fails
- [ ] Show an error when no track is found in search
- [X] Render bathimetry in default basemap
- [x] Don't render any component until login is done
- [x] Fix vessel-sprite URL path
- [x] Clean input after a successful search
- [x] Debug why cursor doesn’t change when hovering clickable features


j8seangel:
@satellitestudiodesign merging to fix the broken things in the deployment, and then we can fix these two things later in a new PR:
- [ ] Log in again when a fetch fails
- [ ] Show an error when no track is found in search

PR #2125, opened 2023-07-07
Refactor analytics hook into a common lib and reuse it on VV, Map and Api Portal. ENG-516

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/ENG-516


apps/api-portal/app/analytics.hooks.ts
```diff
@@ -13,100 +13,20 @@ export enum TrackCategory {
...
-        user_organization: '',
-        user_language: '',
-      })
-    }
-  }, [user, trackLogin, config.length, isLoading])
+  useAnalyticsBase({
+    debugMode: GOOGLE_ANALYTICS_DEBUG_MODE,
+    googleMeasurementId: GOOGLE_MEASUREMENT_ID,
+    googleTagManagerId: GOOGLE_TAG_MANAGER_ID,
+    gtagUrl: 'https://www.googletagmanager.com/gtm.js',
```
j8seangel, line 27: As we have already [this as default value](https://github.com/GlobalFishingWatch/frontend/pull/2125/files#diff-f0c065649cf62e5986f504fcbe281d0bcd480c3a2d078d228c5391956c2c8e14R53) what do you think of removing in the apps and passing only when a different URL is needed?

libs/react-hooks/src/use-analytics/use-analytics.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export type TrackEventParams<T> = {
+  category: T | TrackCategory
+  action: string
+  label?: string
+  value?: any
+}
+
+export const trackEvent = <T>({ category, action, label, value }: TrackEventParams<T>) => {
```
j8seangel, line 18: 👏 love the way of reusing generics [here](https://github.com/GlobalFishingWatch/frontend/pull/2125/files#diff-21064761a7704a322805ab08d7ef25cfd10b406974c3a5a2862a4452941ed1f3R27)

apps/fishing-map/features/app/analytics.hooks.ts
```diff
@@ -26,72 +26,21 @@ export enum TrackCategory {
...
-        action: 'Login',
-      })
-    }
-  }, [user])
+  useAnalyticsBase({
+    debugMode: GOOGLE_ANALYTICS_DEBUG_MODE,
+    googleMeasurementId: GOOGLE_MEASUREMENT_ID,
+    googleTagManagerId: GOOGLE_TAG_MANAGER_ID,
+    gtagUrl: 'https://www.googletagmanager.com/gtm.js',
+    isLoading: false,
```
j8seangel, line 41: as always is false does it worth it to pass this value?

apps/api-portal/app/analytics.hooks.ts
rdgfuentes: Sure, it makes sense.

apps/fishing-map/features/user/user.slice.ts
```diff
@@ -132,4 +132,10 @@ export const isGuestUser = createSelector([selectUserData], (userData) => {
   return userData?.type === GUEST_USER_TYPE
 })
 
+export const isUserLogged = createSelector(
```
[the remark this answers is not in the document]
rdgfuentes: Right, it was causing an error at build time.

PR #2128, opened 2023-07-10
Deck migration/mapbox dependencies clean up

weberjavi:
### Map Controls
- [x] Consume map loaded state from MapControls (derive it from individual layers loaded state)
  - [x] 4wings layer loaded state
  - [x] Vessels layer loaded state
  - [ ] Context layer loaded state
  - [ ] Base layer loaded state
- [x] Zoom buttons
- [x] Miniglobe (bounding box)
- [x] Search button (fitBounds)
- [x] Screen capture button
- [ ] Measure button

### Sidebar
- [ ] Activity/Environmental legends
- [ ] Features in screen for [context layers](https://github.com/GlobalFishingWatch/frontend/blob/61001e14d1cc214aa93bc0af51ebc811b324ad34/apps/fishing-map/features/workspace/context-areas/ContextAreaLayerPanel.tsx#L77)
- [ ] Slider range [histogram](https://github.com/GlobalFishingWatch/frontend/blob/61001e14d1cc214aa93bc0af51ebc811b324ad34/apps/fishing-map/features/workspace/environmental/HistogramRangeFilter.tsx#L41)

### URL state
- [x] [Sync atom and url](https://github.com/GlobalFishingWatch/frontend/blob/61001e14d1cc214aa93bc0af51ebc811b324ad34/apps/fishing-map/features/map/map-viewport.hooks.ts#L29) with map coordinates (feel free to replace with jotai if you prefer to normalize the atoms use)

### Reports
- [ ] Replace [useFilteredTimeSeries ](https://github.com/GlobalFishingWatch/frontend/blob/61001e14d1cc214aa93bc0af51ebc811b324ad34/apps/fishing-map/features/reports/reports-timeseries.hooks.ts#L84) reading data from Deck instances


apps/fishing-map/features/map/popups/TileClusterLayers.tsx
```diff
@@ -94,7 +94,7 @@ function EncounterTooltipRow({ feature, showFeaturesDetails }: EncountersLayerPr
 
   const event = parseEvent(feature.event)
   const linkParams = {
-    ...viewport,
+    ...viewState,
```
j8seangel, line 97: This is going to include a lot of params that we don't need (like maxPitch, minPitch,minZoom,normalize,pitch...) when we only need latitude, longitude, and zoom.
What do you think of having a selector in the atom to return just these three properties?

apps/fishing-map/features/app/App.tsx
```diff
@@ -197,7 +197,7 @@ function App() {
       if (reportAreaBounds) {
         fitMapBounds(reportAreaBounds, { padding: FIT_BOUNDS_REPORT_PADDING })
       } else {
-        setMapCoordinates({ latitude: 0, longitude: 0, zoom: 0 })
+        setViewState({ ...viewState, latitude: 0, longitude: 0, zoom: 0 })
```
j8seangel: I'd use the [useSetAtom](https://jotai.org/docs/core/use-atom#usesetatom) to avoid re-rendering the entire app on every viewState change:
```
This is especially useful when the performance is a concern, as the const [, setValue] = useAtom(valueAtom) will cause unnecessary rerenders on each valueAtom update.
```
See [this commit](https://github.com/GlobalFishingWatch/frontend/commit/24fda5c4eac0ae0d99845520902a08549a1ebc68) as an example of how to avoid it

j8seangel (commented):
I couldn't use the map properly as the zoom got stuck, and I had an infinite loop.
image

I'm going to debug it and will let you know once fixed in a new PR with some proposals

PR #2132, opened 2023-07-14
Track difference between messages table. MAP-1207

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/MAP-1207


j8seangel (approved):
👏 thanks for all the tests! I've normalized the luxon version and moved to peerDependency, so they are not included in the bundle.
Let's talk when you are back about why [this](https://github.com/GlobalFishingWatch/frontend/pull/2132/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R13) was needed but I'm going to merge this to get it fixed

PR #2135, opened 2023-07-18
add track footprint to basic search results

satellitestudiodesign:
- [X] Add map and track
- [X] Request track only when the result is visible on screen
- [X] Request tracks as binary to avoid eating into user quota
- [x] Highlight segments based on TransmissionsTimeline hover
- [x] Show "something" when track response is empty


apps/fishing-map/features/search/TrackFootprint.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const onScreen = useOnScreen(fullCanvasRef)
+  const fullContext = fullCanvasRef.current?.getContext('2d')
+  const highlightContext = highlightCanvasRef.current?.getContext('2d')
+
+  const fetchData = async (vesselId: string) => {
+    if (!trackDatasetId) {
+      setError(true)
+      return
+    }
+    let data = await GFWAPI.fetch<any>(
```
j8seangel: At some point, we should resolve this URL with the `resolveEndpoint` function, but this is enough for now to start testing

apps/fishing-map/hooks/screen.hooks.ts
```diff
@@ -110,3 +110,28 @@ export const useDownloadDomElementAsImage = (
     previewImageLoading,
   }
 }
+
+export const useOnScreen = (ref, rootMargin = '0px') => {
```
j8seangel, line 114: I'd say to move this to the reusable react-hooks library, as it looks very useful!

libs/react-hooks/src/use-small-screen/use-small-screen.ts
```diff
@@ -3,7 +3,7 @@ import { useState, useEffect, useCallback } from 'react'
 const DEFAULT_BREAKPOINT = 768
 
 export function useSmallScreen(width = DEFAULT_BREAKPOINT) {
-  const [isSmallScreen, setIsSmallScreen] = useState<boolean>(false)
+  const [isSmallScreen, setIsSmallScreen] = useState<boolean>(true)
```
j8seangel, line 6: Isn't this a breaking change?

j8seangel (approved):
This is amazing 🥇 I love it!
I've left some minor comments and pushed a [commit](https://github.com/GlobalFishingWatch/frontend/commit/79d4e030faad468e71aa154a65720d9186de9056) with the reviewed changes.

libs/react-hooks/src/use-small-screen/use-small-screen.ts
satellitestudiodesign: I checked where it was used and looked like `true` made more sense as the initial value in all cases. In some places it makes no difference but in others it avoids an unnecessary layout shift. We can review it together.

PR #2136, opened 2023-07-19
SAR - Fixed infrastructure

weberjavi:
[Designs](https://projects.invisionapp.com/d/main#/console/22328119/473974974/preview)

Pending things:
- [x] Fix tooltip content for buffered area
- [x] Add highlighted style for buffered area
- [x] Disable confirm buffer when no buffered area is available (in small areas with large negative values)

Bugs:
- [x] Removing the buffer doesn't request a new report

Blockers:
- [ ] Api returns timeout when requesting a point area like [this one](https://gateway.api.dev.globalfishingwatch.org/v2/datasets/public-unknown-infrastructure-1693471411496/user-context-layer-v1/1536)


apps/fishing-map/features/reports/title/ReportTitle.tsx
```diff
@@ -19,25 +30,73 @@ type ReportTitleProps = {
...
+  const { dispatchQueryParams } = useLocationConnect()
+  const dispatch = useAppDispatch()
   const areaDataview = useSelector(selectReportAreaDataview)
   const report = useSelector(selectCurrentReport)
+  const urlBufferValue = useSelector(selectUrlBufferValueQuery)
+  const urlBufferUnit = useSelector(selectUrlBufferUnitQuery)
+  const fitBounds = useMapFitBounds()
+  const { cleanFeatureState } = useFeatureState(useMapInstance())
+
+  const [bufferValue, setBufferValue] = useState<number>(urlBufferValue || DEFAULT_BUFFER_VALUE)
```
j8seangel, line 42: Is this state needed for performance? Or could we use directly the previewBuffer in redux?

apps/fishing-map/public/locales/source/translations_old.json
j8seangel, line 1: This will need to be reviewed before merging in case we remove needed keys

j8seangel (commented):
It looks great! 🥇 
I left some comments about the code and opening the conversation about a couple of ~bugs~ functionalities:

- What do you think of rendering the preview area just when the tooltip opens? because It is not rendered until yo move the slider
image

- When the title is large or the window small the buttons looks weird, could you take a look?
image

- The buffered area doesn't have any tooltip and I think it is a big confusing, should we have something like "[X] kilometers buffered area from [area]" cc @satellitestudiodesign ?

apps/fishing-map/features/reports/reports.selectors.ts
```diff
@@ -447,3 +455,58 @@ export const selectReportVesselsGraphDataOthers = createSelector(
...
+    return getBufferedFeature({ area, value, unit })
+  }
+)
+
+export const selectReportBufferArea = createSelector(
+  [selectReportAreaData, selectUrlBufferUnitQuery, selectUrlBufferValueQuery],
+  (area, unit, value) => {
+    if (!area || !unit || !value) return null
+    const bufferedArea = getBufferedArea({ area, value, unit }) as Area
+    if (bufferedArea?.bounds && bufferedArea?.geometry) {
```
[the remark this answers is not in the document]
weberjavi: Nope, you're right (legacy from a previous version) 🙇🏼

apps/fishing-map/features/reports/title/ReportTitle.tsx
weberjavi: 🔩

PR #2141, opened 2023-07-21
test remove merge vessel different dataset

fertandil87:


fertandil87:
@rdgfuentes if I am not wrong the events endpoint don't use datasets, I think this restriction was made for tracks

fertandil87:
You can try the example of leah where she reported the problem, and removing this looks good, and the datasets are different

PR #2149, opened 2023-07-27
Fishing map/search and vessel profile

j8seangel:
Search + Vessel Profile using identity dataset


apps/fishing-map/features/dataviews/dataviews.slice.ts
```diff
@@ -388,6 +388,11 @@ export const selectActiveVesselsDataviews = createDeepEqualSelector(
   (dataviews) => dataviews?.filter((d) => d.config?.visible)
 )
 
+export const selectVesselProfileDataview = createDeepEqualSelector(
+  [selectActiveVesselsDataviews, selectVesselInfoData],
```
j8seangel, line 392: We can grab the vessel id from the URL so we don't need to wait the API info response to resolve the dataview

```suggestion
  [selectActiveVesselsDataviews, selectVesselId],
```

PR #2176, opened 2023-08-12
Million research

j8seangel:
https://million.dev

Testing URL: https://fishing-map-fishing-map-million-jzzp2ui3wq-uc.a.run.app/map


j8seangel:
Closing this because there is not much difference as the DOM changes we have are not noticeable.

PR #2228, opened 2023-09-19
fix tests for staging

fertandil87:


apps/fishing-map-e2e/src/support/commands.ts
```diff
@@ -74,7 +74,7 @@ function loginViaAuthAPI(username: string, password: string) {
     const token = interception.response.body.token
     cy.getAllLocalStorage().then((result) => {
       expect(result).to.deep.contain({
-        [Cypress.config('baseUrl')]: {
+        [Cypress.config('baseUrl').replace('/map', '')]: {
```
rdgfuentes, line 77: Good catch!

apps/fishing-map/features/footer/Footer.tsx
```diff
@@ -45,7 +45,7 @@ const FooterPartners = ({ smallScreen }: FooterPartnersProps) => {
           <a href="https://skytruth.org/" rel="noopener noreferrer" target="_blank">
             <img src={LogoSkytruth.src} alt="Skytruth" width="79px" height="24px" />
           </a>
-          <a href="https://google.com/" rel="noopener noreferrer" target="_blank">
+          <a href="https://google.com" rel="noopener noreferrer" target="_blank">
```
rdgfuentes, line 48: Revert this change 🧹  if it's not needed

apps/fishing-map-e2e/src/e2e/map/analysis.cy.ts
```diff
@@ -34,10 +34,10 @@ describe('See the creation of analysis for an area', () => {
   //MAP-1218
   it('Should create an analysis for an EEZ area', () => {
     getSidebar().findBySelLike('report-title').contains(SEARCH_EEZ_FULL_NAME)
-    cy.getBySel('source-tags').findBySelLike('source-tag-item').contains('AIS')
+    cy.getBySel('source-tags', { timeout: 10000 }).findBySelLike('source-tag-item').contains('AIS')
```
rdgfuentes, line 37: IMHO using specific arbitrary timeouts for elements looks similar than adding [Unnecessary `wait()` statements](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting), which is an anti-pattern.
Instead I'd try to setup an approach to adjust the [Timeouts](https://docs.cypress.io/guides/references/configuration#Timeouts) depending on the environment where the e2e tests are running. Eg: We could define  `cypress.config.ts`  files per environment, like `cypress.dev.config.ts` and `cypress.prod.config.ts`, etc... and then set the `defaultCommandTimeout` (or the timeout value we need to set for that environment accordingly. Other could be passing env variables in the cloudbuild step with the proper timeouts we want to use per environment.

fertandil87: no @rdgfuentes , is so different the documentation talks about waits of a number without do anything that means it will sleep 10 seconds and then continue, in this case will give until 10 seconds to appear and is the same that is in lot of places of the tests for api requests and for DOM updates.
and in second place you said that is anti-pattern add it in one place but you like to add it for all?... btw I plan to do that but in another PR, now I am fixing the actual problem

apps/fishing-map/features/footer/Footer.tsx
fertandil87: I needed to have a change in the fishing map or the deploy is not working

j8seangel (commented):
Shouldn't this change go to develop as the base branch first? otherwise, we are going to have the branches unsynced.
And what is the reason this only fails in staging ?

apps/fishing-map-e2e/src/e2e/map/analysis.cy.ts
rdgfuentes: Right, that's why I said similar, it's not exactly the same but setting arbitrary timeouts per elements doesn't sound good to me and it goes in the same direction of the unnecessary waiting, it makes noise. Why using 10 seconds for some elements and 20 for others? I consider this is not scalable and harder to maintain. That's why I think it would be better to tweak the different types of timeouts per environment if necessary. Or improve the app to respond faster so that the default timeouts are not reached. WDYT @j8seangel ?

j8seangel: @rdgfuentes the faster we respond, the better, for sure! do you know what time or machines are running the tests
?
And I think there is some confusion in the discussion between waiting for something and setting a timeout.
Using timeouts it is not: `it goes in the same direction of the unnecessary waiting` because if it finishes we won't wait.
Anyway I agree with having the different configurations in different files, we can do it in another PR

j8seangel:
I'm going to merge to test this in the preview branch and make the test valid again, but let's open a new one for the timeouts configuration

PR #2238, opened 2023-09-21
Fishing map/search and vessel profile terminology

j8seangel:
This includes support for HTML rich content in data terminology
image


satellitestudiodesign (approved):
Looking great, added some styling 💅

PR #2265, opened 2023-10-06
Fishing map e2e/refactor timeouts

fertandil87:
Fixed tests and added custom timeouts


apps/fishing-map-e2e/src/support/app.po.ts
```diff
@@ -36,6 +36,7 @@ export const disablePopups = () => {
   localStorage.setItem('DisableWelcomePopup', 'true')
   localStorage.setItem('DisableSourceSwitchPopup', 'true')
   localStorage.setItem('HighlightPopup', '"vms-with-png"')
+  localStorage.setItem('WelcomePopup', '{"visible":false,"showAgain":false}')
```
j8seangel, line 39: We have a new popup for the vessel profile and changed the behaviour of the marine manager to follow the same logic.

j8seangel (approved):
Apart from the comment of the welcome popups, feel free to merge once fixed!

apps/fishing-map-e2e/src/support/map/eez-download.po.ts
```diff
@@ -71,7 +77,8 @@ export const testCsvOptions = (
 ) => {
   cy.log(description)
   deleteDownloadsFolder()
-  buttons.forEach((button) => cy.getBySel(container).findBySel(button).click())
+  // We need force because the buttons are covered by the active element
+  buttons.forEach((button) => cy.getBySel(container).findBySel(button).click({ force: true }))
```
rdgfuentes, line 81: If this is used in other places I'd extract it to a custom command as suggested [here](https://stackoverflow.com/a/61325254)
```node
Cypress.Commands.add('forceClick', {prevSubject: 'element'}, (subject, options) => {
  cy.wrap(subject).click({force: true})
});
```
What do you think?

apps/fishing-map-e2e/src/e2e/sidebar/search-vessel.cy.ts
```diff
@@ -36,22 +41,27 @@ xdescribe('Basic search for a vessel', () => {
...
         .then((vessel) => {
           // Click on the first result
-          cy.getBySel('search-vessels-list').findBySelLike('search-vessels-option').eq(0).click()
+          cy.getBySel('search-vessels-list')
+            .findBySelLike('search-vessels-option-selection-0')
+            .click()
           cy.getBySel('search-vessels-add-vessel').click()
           scrollSidebar('center', 2000)
+          console.log(vessel)
+          console.log(vessel.text())
```
rdgfuentes, line 62: Remove these console.logs if not needed 🧹  or use `cy.log` instead.

apps/fishing-map-e2e/src/support/app.po.ts
fertandil87: ok good to know, but it is not necessary right know, I will add it when a test in those pages appear

PR #2283, opened 2023-10-18
Sar point buffer fixes

weberjavi:
This PR keeps a record of the original geometry type of the buffered feature, before dissolving it, to allow the UI to behave differently in case the original geometry is a point.


weberjavi:
@j8seangel this is ready for review. Will handle `highlight` issues in a different PR.

PR #2310, opened 2023-10-23
Fishing map/v3 integration fixes

weberjavi:
This PR fixes two small issues. 
-  The first one is related with API v3 updates. [`/v3/4wings/report` ](https://gateway.api.dev.globalfishingwatch.org/swagger?version=3#/4wings/get_v3_4wings_report) endpoint need params to be Uppercase (a `422` was being returned).
- Cleaning feature state on report buffer operation update was caussing the issue of leaving no area highlighted in case the buffer was removed before confirmation.


j8seangel (approved):
💯

PR #2339, opened 2023-10-27
Fishing map/more buffer params cleanup

weberjavi:
This PR fixes a couple of issues related to buffer state:
- When you open the buffer area tooltip (coming from a saved report) the default values doesn't match the buffer selection
- The area is not clean if you remove it.

Some thinking might be needed on the title creation logic (when the buffer is removed on a saved report with buffer the name of the reports remains the same but the report area is the original one, letting the name kind of out of sync) 
c.c: @satellitestudiodesign


j8seangel (approved):
👌 we can make the decision of the report title in another PR, thanks!

PR #2363, opened 2023-11-07
fix loading legacy track data

j8seangel:
Support loading tracks from different dataset than the configured in the dataview datasetConfig
Pending confirmation for the same with event datasets


rdgfuentes (approved):
LGTM

PR #2364, opened 2023-11-07
Fishing map/fix profile permissions

j8seangel:
https://globalfishingwatch.atlassian.net/browse/MAP-1258

image


apps/fishing-map/features/user/UserLoggedIconButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <IconButton
+          {...props}
+          icon={isLoginHover ? 'user' : props.icon}
+          tooltip={
+            props.loginTooltip ||
+            t('vessel.infoLogin', 'Register and login to see more details (free, 2 minutes)')
+          }
+          onMouseEnter={() => setIsLoginHover(true)}
+          onMouseLeave={() => setIsLoginHover(false)}
+        />
```
weberjavi, line 27: We would need to avoid passing `onClick` handler to the `IconButton` to avoid the download to be executed when clicking (and only triggering the login link action). Maybe passing undefined as `onClick` value?

 ```suggestion
          onClick={undefined}
        />
```

weberjavi (changes requested):
Button is triggering the login flow but also downloading the data (since we are passing the `onClick` prop to the child `IconButton`)

apps/fishing-map/features/user/UserLoggedIconButton.tsx
j8seangel: Great catch! fixed [here](https://github.com/GlobalFishingWatch/frontend/pull/2364/commits/11e6ddea1960cd67b8dbb5a717d9b2f6df2847d2)

weberjavi (approved):
👍🏼

PR #2365, opened 2023-11-07
Ship name on popup

weberjavi:
This PR updates the way the `shipname` is resolved on the events popup to handle several identities.


j8seangel (approved):
👌

PR #2373, opened 2023-11-10
skip unneeded build step

j8seangel:
While I was testing to build libraries using [swc](https://nx.dev/nx-api/js/executors/swc#nxjsswc) for faster builds I realized the [incremental builds](https://nx.dev/concepts/more-concepts/incremental-builds#when-should-i-use-incremental-builds) are not working right now.

The own app is compiling the dependencies as `transpilePackages` in next so basically we are spending time on building libraries that we are not used.

This PR disables that step until we find a better solution, @weberjavi @rdgfuentes @satellitestudiodesign  could you test the `nx build [app]` works for you too?

### DIFF
--- a/libs/api-client/project.json
+++ b/libs/api-client/project.json
@@ -4,7 +4,7 @@
   "sourceRoot": "libs/api-client/src",
   "projectType": "library",
   "targets": {
-    "build": {
+    "dist": {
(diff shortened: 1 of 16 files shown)


weberjavi:
Working on my end 👌🏼

PR #2379, opened 2023-11-14
Fishing map/timerange reset to one on disabled

weberjavi:
This PR brings back the functionality of reseting lower time intervals to one (month and/or day) when an interval change is forced due to range limit between and start and end date is surpased.

[Grabación de pantalla desde 14-11-23 13:19:43.webm](https://github.com/GlobalFishingWatch/frontend/assets/6906348/54b7374e-38b4-41ed-8fba-61c7ca336c31)


weberjavi:
⏳ everything working
Time might byte us again... but that would be in the future

PR #2396, opened 2023-11-16
Improve debug of the tests - Set production url to the test when we deploy to prod // ENG-564

fertandil87:
https://globalfishingwatch.atlassian.net/browse/ENG-564

Set production url to the test when we deploy to prod


rdgfuentes (approved):
LGTM

PR #2403, opened 2023-11-21
Fishing map/new dataset upload

j8seangel:
New datasets upload workflow

- [ ] Quick drag and drop overlay with 3 drop zones
- [ ] Generate the minimum dataset schema to allow future edits (only supported in new datasets)
- [ ] Add field selections to dataset entity (e.g. pointSize, pointLabel)
- [ ] Add field selections in tile requests (properties)
- [ ] New upload Popup with the 3 options
- [ ] Polygons Upload/edit UI
  - [ ] Optional opacity by numerical field
  - [ ] Optional label by any field
  - [ ] Optional temporal filter by date field/fields
  - [ ] Optional filters (range or string) add to fieldsAllowed
- [ ] Tracks Upload/edit UI
  - [ ] Optional track segment id by any field
  - [ ] Optional filters (range or string) add to fieldsAllowed
- [ ] Points Upload/edit UI
  - [ ] Optional label by any field
  - [ ] Optional point size by numerical field
  - [ ] Optional temporal filter by date field/fields
  - [ ] Optional filters (range or string) add to fieldsAllowed


apps/fishing-map/features/datasets/datasets.utils.ts
```diff
@@ -393,6 +395,24 @@ export const getEventsDatasetsInDataview = (dataview: UrlDataviewInstance) => {
   })
 }
 
+export type DatasetSchemaGeneratorProps = {
+  data: CSV
+  meta: ParseMeta
+}
+export const getDatasetSchema = ({ data, meta }: DatasetSchemaGeneratorProps) => {
```
j8seangel, line 402: As the generic dataset.utils file is big enough, so what about moving all the custom upload functions to its own file within the `upload` folder?

Also, I think the signature of this should be something like `getDatasetSchemaFromCSV`

apps/fishing-map/features/datasets/upload/NewDataset.tsx
```diff
@@ -38,12 +32,11 @@ export type DatasetMetadata = {
   public: boolean
   category: DatasetCategory
```
j8seangel, line 33:
```suggestion
  category: Dataset['category']
```

j8seangel: and the same for the rest of the properties

apps/fishing-map/features/datasets/upload/NewDataset.tsx
```diff
@@ -272,10 +265,12 @@ function NewDataset(): React.ReactElement {
   const onConfirmClick = useCallback(
     async (dataset: DatasetMetadata, datasetFile?: File) => {
       if (dataset) {
-        const { fields, ...meta } = dataset as DatasetMetadata
         const { payload, error: createDatasetError } = await dispatchUpsertDataset({
           dataset: {
-            ...meta,
+            ...dataset,
+            fieldsAllowed: dataset.configuration?.configurationUI?.filter
```
j8seangel, line 271: is needed to duplicate the state here? I mean, why not using directly the `fieldsAllowed` property in the dataset instead of the `filter`

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
```diff
@@ -78,35 +81,29 @@ function NewTrackDataset({
     if (file) {
       updateFileData(file)
     } else if (dataset) {
-      console.log('TODO', dataset)
-      setDatasetMetadata((meta) => ({
-        ...meta,
+      setDatasetMetadata({
         id: dataset.id,
         name: dataset.name,
         public: true,
```
j8seangel, line 87: I think this should be taken from the dataset name instead of hardcoded, could use the existing `isPrivateDataset` function

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
```diff
@@ -78,35 +81,29 @@ function NewTrackDataset({
...
-      }))
-      // setDatasetMetadata(dataset)
+      })
     }
   }, [dataset, file, updateFileData])
 
   const onConfirmClick = useCallback(() => {
     let file: File | undefined
     if (datasetMetadata) {
+      const config = {
```
j8seangel, line 101: 👏 great idea! this way we can make it retro compatible, I think I'd move this to the `upload/datasets-upload.utils` file

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
```diff
@@ -156,22 +153,31 @@ function NewTrackDataset({
...
+        Object.keys(datasetMetadata.schema).map((field) => ({ id: field, label: field }))) ||
+      [],
     [datasetMetadata]
   )
 
   const selectedOption = useCallback(
     (option: string): SelectOption => ({
-      label: datasetMetadata?.configuration?.[option] as string,
-      id: datasetMetadata?.configuration?.[option],
+      label: (datasetMetadata?.configuration?.configurationUI?.[option] ||
```
j8seangel, line 176: Instead of generating a new option with its possible error or keeping it unsynced with the original options, why not do something like:
```
fieldsOptions.find(o => o.id === option)
```

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
```diff
@@ -156,22 +153,31 @@ function NewTrackDataset({
...
       }))
     },
     []
   )
 
   const fieldsOptions: SelectOption[] = useMemo(
-    () => datasetMetadata?.fields?.map((field) => ({ id: field, label: field })) || [],
+    () =>
+      (datasetMetadata?.schema &&
+        Object.keys(datasetMetadata.schema).map((field) => ({ id: field, label: field }))) ||
```
j8seangel, line 169: a very very nice to have feature could be to remove from this list the fields already selected

j8seangel (commented):
Some comments, but this is going great!

apps/fishing-map/features/datasets/upload/NewDataset.tsx
weberjavi: We need the newly set `filter` to update the fieldsAllowed, if we do not have `filter` then yes, we would need to use already set dataset `filedAllowed`

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
weberjavi: But as far as I understand it what we want to display as `selectedOption` is the one that is currently selected in the dataset configuration

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
weberjavi: What should be the value in newly created datasets?

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
weberjavi: Wouldn't we eventually want add more than one property to the filters array? I mean to add and remove items from the fields list? So we would need to have all available to toggle them on the list

apps/fishing-map/features/datasets/upload/NewTrackDataset.tsx
j8seangel: It is public by default

PR #2405, opened 2023-11-23
Fishing map/new dataset upload gdal

j8seangel:
Trying https://github.com/bugra9/gdal3.js to read and convert dataset to upload


j8seangel:
Apart of the complexity of extracting dataset metadata using SQL queries the .wasm + .data file weights around 48MB so discarding this approach as the bundle size is not affordable

PR #2417, opened 2023-11-30
VV-577. VV Link to VV 2.0 in Map with no fishing effort layer enabled

rdgfuentes:
https://globalfishingwatch.atlassian.net/browse/VV-577


j8seangel (approved):
👌 thanks for the comment! it will help to remember why

PR #2420, opened 2023-12-11
Fishing map/dataset upload improvements

satellitestudiodesign:


apps/fishing-map/public/locales/source/translations.json
```diff
@@ -256,6 +256,34 @@
     "uploadPublic": "Allow other users to see this dataset when you share a workspace",
     "uploadYourOwn": "Upload your own"
   },
+  "datasetUploadUI": {
```
j8seangel, line 259: Everything here is related to ui. Is that adding any value? or why not just datasetUpload?

PR #2421, opened 2023-12-11
Fishing map/annotations

j8seangel:
🎁


apps/fishing-map/features/map/annotations/Annotations.tsx
```diff
@@ -0,0 +1,85 @@
...
+      closeButton={true}
+      closeOnClick={false}
+      onClose={resetMapAnnotation}
+      maxWidth="330px"
+      className={cx(styles.popup)}
+    >
+      <div className={styles.popupContent}>
+        <div className={styles.flex}>
+          <InputText
+            value={mapAnnotation.label}
```
weberjavi, line 60: Not sure if storing the `annotation.label` on the URL could eventually cause any issue due to too long URLs

weberjavi (approved):
Left some comments that do not avoid the PR to be merged
Really superb christmas gift!! 🎁 🚀

apps/fishing-map/features/map/annotations/Annotations.tsx
j8seangel: it could, but we don't have any other options if we want to make them shareable

PR #2423, opened 2023-12-13
Fishing map/upload tracks and points leftovers

weberjavi:


libs/layer-composer/src/generators/user-points/user-points.ts
```diff
@@ -51,17 +51,17 @@ class UserPointsGenerator {
       'source-layer': DEFAULT_CONTEXT_SOURCE_LAYER,
     }
     let filters: FilterSpecification | undefined
-    if (config?.pointTimeFilterProperty) {
+    if (config?.startTimeFilterProperty && config?.endTimeFilterProperty) {
```
j8seangel, line 54: Does this keep filtering functionality by a timestamp instead of a range?

apps/fishing-map/features/reports/reports.hooks.ts
```diff
@@ -91,7 +90,7 @@ export function useFitAreaInViewport() {
 export const COARSE_SIMPLIFY_TOLERANCE = 0.1
 export const FINE_SIMPLIFY_TOLERANCE = 0.001
 
-export function getSimplificationByDataview(dataview: UrlDataviewInstance | Dataview) {
+export function getSimplificationByDataview(dataview: UrlDataviewInstance | Dataview | undefined) {
```
j8seangel, line 93: I'd rollback this change to ensure there is always a Dataview

apps/fishing-map/features/datasets/upload/NewDatasetField.tsx
```diff
@@ -62,7 +62,7 @@ export function NewDatasetField({
     <InputText
       value={getDatasetConfigurationProperty({
         dataset: datasetMetadata,
-        property: 'timestamp',
+        property,
```
j8seangel, line 65: shame on me, sorry!

libs/layer-composer/src/generators/user-points/user-points.ts
weberjavi: When we filter by `timestamp` what we are doing is setting the start and end time to be the same number, so yeah, it works both for `timestamp` and `timerange`

apps/fishing-map/features/reports/reports.hooks.ts
weberjavi: 👌🏼

PR #2446, opened 2024-01-09
update to redux-toolkit 2

j8seangel:
- [x] Updates redux-toolkit version
- [x] Fixes `The result function returned its own inputs without modification` except
- [x] Fixes `createAsyncSlice` types
- [x] Improve selector typings to [auto-infer values instead of need generics](https://github.com/GlobalFishingWatch/frontend/commit/304b59ded06db5c37298e6752febfb8643565622#diff-7204386cda019914e73013777bac43a6761d8ec5671e3e994df3895bec6b176cR14)


satellitestudiodesign:
There are other selectors that through the same warning (e.g. `selectActiveTrackDataviews`, `selectVesselProfileStateProperty`). All of them seem to have in common that they are (or they use) a function that return a createSelector, so we may still be doing something wrong..

PR #2448, opened 2024-01-12
Fishing map/bathymetry env layer

satellitestudiodesign:
- [x] Use custom reversed color ramps
- [x] Don't allow to change layer color
- [x] Make it always the last layer of its section
- [x] Don't show it in the time-bar selector


j8seangel:
💯

PR #2453, opened 2024-01-17
tasks by 1️⃣ 🚗

satellitestudiodesign:
- [x] fix the type of the file is not compatible for kml/kmz in polygons
- [x] fix the transparent popup bug
- [x] Add explanation tooltips to fields
- [x] Review why filters cursor looks disabled but it allows to select
- [x] Don’t allow to edit timestamp property in uploaded csv track

Additional stuff
- Make "areas on the map" block collapsable (to avoid an unnecessary long scroll that can hinder uploaded datasets section visibility)
Screenshot 2024-01-17 at 12 11 02
Screenshot 2024-01-17 at 12 11 11
- Tweak Time filter selection to make clearer the distinction between selecting a filter type and a dataset field
Screenshot 2024-01-17 at 12 12 28

### DIFF
--- a/apps/fishing-map/features/app/drag-dataset.hooks.ts
+++ b/apps/fishing-map/features/app/drag-dataset.hooks.ts
@@ -42,7 +42,7 @@ export function useDatasetDrag() {
     [dispatchDatasetModalConfig, dispatchDatasetModalOpen]
   )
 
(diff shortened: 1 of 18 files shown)


j8seangel:
💯

PR #2461, opened 2024-01-23
Avoid history replacement on search link click

weberjavi:
This PR avoids the situation in which a pinned vessel is removed from the map when the user performs a new vessel search from the search icon on the sidebar category tabs 👇🏼 
image

This does not happens when the search is performed from the vessel sidebar panel


satellitestudiodesign (approved):
Looks good :)

PR #2545, opened 2024-02-27
deck-migration/annotation-layer

weberjavi:
Annotations and Error layers (they are built with [nebulaGl html overlays](https://nebula.gl/docs/api-reference/overlays/html-overlay), make sure to `yarn` to install needed dependencies).
This Pr also migrate viewport hooks to Jotai.


apps/fishing-map/features/map/annotations/Annotations.tsx
```diff
@@ -1,93 +1,68 @@
...
+  const { upsertMapAnnotations, mapAnnotations, areMapAnnotationsVisible } = useMapAnnotations()
+  const { setMapAnnotation } = useMapAnnotation()
+  const deck = useDeckMap()
+  const handleHover = useCallback(() => {
+    deck?.setProps({ getCursor: () => 'move' })
+  }, [deck])
+  const handleMouseLeave = useCallback(() => {
+    deck?.setProps({ getCursor: () => 'grab' })
+  }, [deck])
+  const handleDragStart = useCallback(() => {
```
j8seangel, line 18: What about hiding the annotation while it is being dragged to avoid this weird effect?

https://github.com/GlobalFishingWatch/frontend/assets/10500650/3d2d6cb8-0b02-4e36-bd1c-bbd3e39ccd44

apps/fishing-map/features/map/annotations/Annotations.tsx
```diff
@@ -1,93 +1,68 @@
...
-  }
-
-  if (!gfwUser) {
-    return null
-  }
+const MapAnnotations = (props: MapAnnotationComponentProps): React.ReactNode | null => {
+  const { viewport } = props
+  const { upsertMapAnnotations, mapAnnotations, areMapAnnotationsVisible } = useMapAnnotations()
+  const { setMapAnnotation } = useMapAnnotation()
+  const deck = useDeckMap()
```
j8seangel, line 11: If we are already reading the deck instance here, why not reading the props needed from it directly? this way we can avoid this [render function](https://github.com/GlobalFishingWatch/frontend/pull/2545/files#diff-f2d803a9498e21b6106db6b639a9374e45cfe24a800cd9f37c3801812dc4a368R313) which should be better in terms of performance. WDTY ?

j8seangel (changes requested):
I'm getting a typing build error [here](https://github.com/GlobalFishingWatch/frontend/blob/deck-migration/annotations-layer/apps/fishing-map/features/map/map-interactions.hooks.ts#L152) can you take a look?
Please take a look at the comments below but in general the nebula integration looks easier that we though! thanks!

weberjavi:
@j8seangel ready again

j8seangel:
I'm still having some small jumps between the positions, but I've added it to the list, so we can improve it later.

https://github.com/GlobalFishingWatch/frontend/assets/10500650/695ec229-c924-42b5-bee3-648839f8b6a4

Merging to avoid having conflicts between the [other big branch](https://github.com/GlobalFishingWatch/frontend/pull/2547)

PR #2560, opened 2024-03-13
Deck migration/fourwings improvements

satellitestudiodesign:
- [x] Bivariate mode
- [x] Fix cache bug when time range includes latest day
- [x] update timebar graph when changing between activity and detections

### DIFF
--- a/libs/dataviews-client/src/merge-dataviews.ts
+++ b/libs/dataviews-client/src/merge-dataviews.ts
@@ -261,21 +261,28 @@ export function getDataviewsMerged(
     )
 
   const singleHeatmapDataview = [...activityDataviews, ...detectionDataviews].length === 1
-  const heatmapAnimatedMode = params.bivariateDataviews
+  const activityHeatmapAnimatedMode = activityDataviews.every((dataview) =>
+    params.bivariateDataviews?.includes(dataview.id)
+  )
+    ? HeatmapAnimatedMode.Bivariate
+    : HeatmapAnimatedMode.Compare
+  const detectionsHeatmapAnimatedMode = detectionDataviews.every((dataview) =>
+    params.bivariateDataviews?.includes(dataview.id)
+  )
     ? HeatmapAnimatedMode.Bivariate
     : HeatmapAnimatedMode.Compare
   // If activity heatmap animated generators found, merge them into one generator with multiple sublayers
   const mergedActivityDataview = activityDataviews?.length
     ? getMergedHeatmapAnimatedDataviews(activityDataviews, {
         ...params,
-        heatmapAnimatedMode,
+        heatmapAnimatedMode: activityHeatmapAnimatedMode,
         colorRampWhiteEnd: singleHeatmapDataview,
       })
     : []
   const mergedDetectionsDataview = detectionDataviews.length
     ? getMergedHeatmapAnimatedDataviews(detectionDataviews, {
         ...params,
-        heatmapAnimatedMode,
+        heatmapAnimatedMode: detectionsHeatmapAnimatedMode,
         colorRampWhiteEnd: singleHeatmapDataview,
       })
     : []
--- a/libs/dataviews-client/src/merge-dataviews.ts
+++ b/libs/dataviews-client/src/merge-dataviews.ts
@@ -91,7 +91,7 @@ type GetMergedHeatmapAnimatedDataviewParams = {
 export function getMergedHeatmapAnimatedDataviews(
   heatmapAnimatedDataviews: UrlDataviewInstance[],
   {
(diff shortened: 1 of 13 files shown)


libs/deck-layers/src/layers/fourwings/FourwingsHeatmapTileLayer.ts
```diff
@@ -235,29 +233,31 @@ export class FourwingsHeatmapTileLayer extends CompositeLayer<
...
   }
 
-  _updateSublayerColorRanges = () => {
-    const { sublayers, colorRampWhiteEnd } = this.props
-    const newSublayerColorRanges = sublayers.map(({ config }) =>
-      HEATMAP_COLOR_RAMPS[
-        (colorRampWhiteEnd ? `${config.colorRamp}_toWhite` : config.colorRamp) as ColorRampsIds
-      ].map((c) => rgbaStringToComponents(c))
-    )
+  _updateStateWithNewProps = () => {
```
j8seangel, line 241: I think we can use the [existing update state method](https://deck.gl/docs/api-reference/core/layer#updatestate) to check if `changeFlags` is `propsChanged` so we can set the new state on props change and avoid one re-render

libs/deck-layers/src/layers/fourwings/FourwingsHeatmapTileLayer.ts
```diff
@@ -73,81 +46,106 @@ export class FourwingsHeatmapTileLayer extends CompositeLayer<
...
-    // console.log('steps:', steps)
-    // console.log('calculateColorDomain time:', fb - fa)
     return steps
   }
 
-  updateColorDomain = debounce((tiles) => {
+  updateColorDomain = debounce(() => {
     requestAnimationFrame(() => {
-      this.setState({ colorDomain: this.calculateColorDomain(tiles) })
+      this.setState({ colorDomain: this.calculateColorDomain(this.state.comparisonMode) })
```
j8seangel: I think would be more clear to have all the needed data in the same format, I mean:
- sending all the data as function params so we can extract the function out of the class
- reading all the data directly from the state

but in this case we send the comparisonMode as a parameter but the tiles from the [state](https://github.com/GlobalFishingWatch/frontend/pull/2560/commits/7d3bb738a2d5fb6207838090cfcef96ca4b8c7f0#diff-019339b86e8de09cbdbbb40dc141be52d4e4c5278f335c9451dbcb8a94d6f34aR91) WDYT ?

libs/deck-layers/src/utils/colorRamps.ts
```diff
@@ -0,0 +1,161 @@
...
+  '#3DADDF', // -50000 -> -100000
+  '#4286C1', // -8000 -> -50000
+  '#3F69A4', // 0 -> -8000
+  '#4A4174',
+  '#844A7F',
+  '#B0547F',
+  '#FF677D',
+]
+
+export const HEATMAP_COLOR_RAMPS: Record<ColorRampsIds, string[]> = {
```
j8seangel, line 140: just a comment to remember that one day we'll be able to remove all of this code

libs/deck-layers/src/layers/fourwings/fourwings.utils.ts
```diff
@@ -215,3 +226,116 @@ export const aggregatePositionsTimeseries = (positions: Feature[]) => {
...
+  const colIndex = (valueA || 1) - 1
+  const rowIndex = (valueB || 1) - 1
+
+  const index = rowIndex * 4 + colIndex
+  // offset by one because values start at 1 (0 reserved for values < min value)
+  return index + 1
+}
+
+const EMPTY_CELL_COLOR: Color = [0, 0, 0, 0]
+export const chooseColor = (
```
j8seangel, line 272: it is great to have this function out of the layer, thanks!

libs/dataviews-client/src/merge-dataviews.ts
```diff
@@ -261,16 +261,21 @@ export function getDataviewsMerged(
...
 
   const singleHeatmapDataview = [...activityDataviews, ...detectionDataviews].length === 1
+  const heatmapAnimatedMode = params.bivariateDataviews
+    ? HeatmapAnimatedMode.Bivariate
+    : HeatmapAnimatedMode.Compare
   // If activity heatmap animated generators found, merge them into one generator with multiple sublayers
   const mergedActivityDataview = activityDataviews?.length
     ? getMergedHeatmapAnimatedDataviews(activityDataviews, {
         ...params,
+        heatmapAnimatedMode,
```
j8seangel, line 271: I have some doubts about how to integrate this because the bivariate mode is not compatible between these two layers. let's review it together tomorrow

j8seangel (commented):
👏 👏 👏 👏

PR #2587, opened 2024-04-02
Normalize layers load state

weberjavi:


j8seangel (approved):

libs/deck-layers/src/layers/context/ContextLayer.ts
```diff
@@ -87,6 +92,7 @@ export class ContextLayer<PropsT = {}> extends CompositeLayer<_ContextLayerProps
             return [
               new GeoJsonLayer(mvtSublayerProps, {
                 id: `${props.id}-boundaries`,
+                onViewportLoad: this._onViewportLoad,
```
j8seangel, line 95: as we are now doing anything custom with the callback now why not removing the `_onViewportLoad` method and use `this.props.onViewportLoad` callback directly?

j8seangel (commented):
Great! looks like it works like a charm without the need of re-implementing the loaded state.

However, there are some some pieces of the previous implementation of the search state, in [fourwings](https://github.com/GlobalFishingWatch/frontend/blob/deck-migration/normalize-layers-load-state/libs/deck-layers/src/layers/fourwings/FourwingsLayer.ts#L34), [context](https://github.com/GlobalFishingWatch/frontend/blob/deck-migration/normalize-layers-load-state/libs/deck-layers/src/layers/basemap/BasemapLayer.ts#L23) and [vessels](https://github.com/GlobalFishingWatch/frontend/blob/deck-migration/normalize-layers-load-state/libs/deck-layers/src/layers/vessel/VesselEventsLayer.ts#L24) could you clean them all? 

thanks!!

PR #2591, opened 2024-04-04
improve 4wings rendering

satellitestudiodesign:
BEFORE
Screenshot 2024-04-09 at 13 11 19

AFTER
Screenshot 2024-04-09 at 13 11 26

### DIFF
--- a/libs/deck-layers/src/layers/fourwings/FourwingsHeatmapTileLayer.ts
+++ b/libs/deck-layers/src/layers/fourwings/FourwingsHeatmapTileLayer.ts
@@ -123,7 +123,7 @@ export class FourwingsHeatmapTileLayer extends CompositeLayer<
         return this.getColorDomain()
       }
 
-      const steps = allValues.map((sublayerValues) =>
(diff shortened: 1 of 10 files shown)


libs/deck-layers/src/layers/fourwings/FourwingsHeatmapTileLayer.ts
```diff
@@ -123,7 +124,10 @@ export class FourwingsHeatmapTileLayer extends CompositeLayer<
       }
 
       const steps = allValues.map((sublayerValues) =>
```
weberjavi, line 126: Screenshot 2024-04-09 at 15 55 05

If any of the sublayers has no data on the current tiles the `ckmeans` calculation crashes the app. Filtering empty sublayers would avoid that situation
```suggestion
      const steps = allValues.filter((sublayer) => sublayer.length).map((sublayerValues) =>
```

weberjavi (changes requested):
Found an issue (it is not related to the changes in this PR though)

PR #2627, opened 2024-04-23
test loading tracks with tiles

j8seangel:


j8seangel:
Not feasible in the api side

PR #2708, opened 2024-06-27
Filterable fixed infrastructure

weberjavi:
This PR adds new filterable fixed infrastructure layer to the layer library.
It also updates the way the `USER_POINTS` layers are filtered in the map.


j8seangel:
https://fishing-map.dev.globalfishingwatch.org/map/fishing-activity/filtered_fixed_infrastructure-public

PR #2856, opened 2024-10-01
Disable events tab for non GFW users

weberjavi:
- [x] Disable events tab for non GFW users
- [x] Allow users to filter activity layers by vessel groups present on the workspace but not theirs
- [x] When a user is logged in and the vessel group is not owned by him it does not request the group (it does when the user is not logged) 
- [x] Update encounter events sub-tab translation string


apps/fishing-map/features/vessel-groups/vessel-groups.slice.ts
```diff
@@ -80,7 +80,7 @@ export const fetchVesselGroupsThunk = createAsyncThunk<
     const vesselGroupsParams = {
       ...DEFAULT_PAGINATION_PARAMS,
       cache: false,
-      'logged-user': true,
+      // 'logged-user': true,
```
j8seangel, line 83: we needed this to avoid admins or gfw-staff requesting all the public vessel-groups in the API, I think we need to get it back but only when needed

PR #2907, opened 2024-10-30
Add resizable functionality

weberjavi:
To check this functionality working add `isResizable={true}` property to `fishing-map` `Timebar` component in `apps/fishing-map/features/timebar/Timebar.tsx`


satellitestudiodesign (approved):
Looks good. My only doubt is with the 200px max height. I'd put something much bigger (like 400) in case users have a big screen and want to take advantage of it when we add the ability to select points in the time bar in the track-labeler.

weberjavi:
@satellitestudiodesign will update the max height and merge.

PR #2931, opened 2024-11-21
Map fixes and timebar interactions

weberjavi:
- [x] Bring back rulers and layers toggling.
- [x] Update timebar points styles depending on filter range values.
- [x] Update new segment styles on sidebar and highlight map track on hover
- [x] Update defaults (timebar height and display mode)


j8seangel (approved):
This is great! 🙇  
I've updated the branch with develop as there were some conflicts after running the updated eslint config

PR #2960, opened 2024-12-12
Add GA script

weberjavi:


j8seangel:
Ideally, we could move the track `ID` to an env variable, but as this is going to be used only internally I think is not that important

PR #2997, opened 2025-01-21
Track labeler/rc slider update

weberjavi:
This PR avoids `rc-slider `to crash the app due to the incompatibility with current (recently updated) `react-dom` version of the project.


j8seangel (approved):
works well and nice to have dependencies up to date 👏

PR #3004, opened 2025-01-24
Responsive visualizations in fishing-map

j8seangel:
Enable individual data render from https://github.com/GlobalFishingWatch/frontend/pull/2988


j8seangel:
Implemented in https://github.com/GlobalFishingWatch/frontend/pull/3031

PR #3089, opened 2025-03-20
info correction modal tweaks

satellitestudiodesign:


apps/fishing-map/features/vessel/infoCorrection/InfoCorrectionModal.tsx
```diff
@@ -139,14 +136,12 @@ function InfoCorrectionModal({ isOpen = false, onClose }: InfoCorrectionModalPro
...
 
-  const onSourceClick = (choice: ChoiceOption<any>) => {
-    dispatchQueryParams({ vesselIdentitySource: choice.id })
-    trackEvent({
-      category: TrackCategory.VesselProfile,
-      action: 'click_info_correction_source',
-      label: String(choice.label),
-    })
-  }
+  const onSourceClick = useCallback(
```
GeorgiaDesigns, line 139: When do we need to track events? Is there someone that requests event tracking in certain functionalities? Because I assumed we were tracking all events

apps/fishing-map/features/vessel/infoCorrection/InfoCorrectionModal.tsx
```diff
@@ -58,8 +59,6 @@ type InfoCorrectionModalProps = {
 }
 
 const VALID_REGISTRY_FIELDS = [
-  'transmissionDateFrom',
-  'transmissionDateTo',
```
GeorgiaDesigns, line 62: Why did you remove the transmission Dates? Is this not something we will allow to be modified?

GeorgiaDesigns (approved):

apps/fishing-map/features/vessel/infoCorrection/InfoCorrectionModal.tsx
satellitestudiodesign: It's usually up to us and Gisela, who manages the Analytics dashboard. In this case, since this is an internal use case, we don't even need to track anything.. but I added it to the "send" action in case we delete corrections in the table and we want to know how many people sent them.. but probably not needed

apps/fishing-map/features/vessel/infoCorrection/InfoCorrectionModal.tsx
satellitestudiodesign: The dates are metadata; we get the information directly from the satellites, so there's no way to change it. There's no way for users to consider them right or wrong.

PR #3102, opened 2025-03-31
Fishing map/new user layers

GeorgiaDesigns:
* Layer library
    * Add user categories
    * Add Vessel group section
    * Add draw a point in user section


j8seangel (changes requested):
I have a weird bug when clicking on the `Events` category as the sidebar marks the environmental section as active. can you take a look?

https://github.com/user-attachments/assets/10c438bb-d701-409b-9a3a-1992ecc99549

Also please take a look at the comments below to cleanup the code

GeorgiaDesigns:
Done, updated! check if there are any more changes needed, but I think everything is working now

GeorgiaDesigns (commented):
review check

PR #3103, opened 2025-03-31
prevent DeckGL from Resizing During Print

GeorgiaDesigns:
Fix for this bug ticket: https://globalfishingwatch.atlassian.net/browse/MAP-1498

Force the Page to Use a Fixed Size for Printing
prevent DeckGL from Resizing During Print

### DIFF
--- a/apps/fishing-map/features/app/App.module.css
+++ b/apps/fishing-map/features/app/App.module.css
@@ -51,8 +51,6 @@
     height: auto;
     min-height: 100%;
     background: none;
-
-    /* width: 21cm; */
     display: flex;
     flex-direction: column-reverse;
     overflow: visible;
@@ -70,6 +68,7 @@
   .main {
     position: relative;
     width: 100%;
+    max-height: 28.5cm;
     left: 0 !important;
     height: 20cm !important;
   }
--- a/apps/fishing-map/features/app/App.module.css
+++ b/apps/fishing-map/features/app/App.module.css
@@ -38,17 +38,21 @@
 
 @page {
   size: 21cm 29.7cm;
-  margin: 1cm;
 }
 
 @media print {
+  body {
+    margin: 1cm;
+  }
+
   .splitContainer {
     --color-primary-blue: black;
 
     height: auto;
     min-height: 100%;
     background: none;
-    width: 21cm;
+
+    /* width: 21cm; */
     display: flex;
     flex-direction: column-reverse;
     overflow: visible;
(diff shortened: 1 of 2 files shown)


apps/fishing-map/features/app/App.module.css
```diff
@@ -66,9 +70,8 @@
   .main {
     position: relative;
     width: 100%;
-    height: 28.5cm;
```
j8seangel, line 69: I think the idea of this value is to ensure every page fits on a A4 page, and right now is not happening
Google Chrome 2025-04-01 10 20 33
do you think we can keep this value and update only the map?

apps/fishing-map/features/map/DeckGLWrapper.tsx
```diff
@@ -124,6 +148,7 @@ const DeckGLWrapper = () => {
       onDrag={onMapDrag}
       onDragEnd={onMapDragEnd}
       onLoad={onMapLoad}
+      useDevicePixels={false}
```
j8seangel, line 151: why is this needed?

j8seangel (changes requested):

apps/fishing-map/features/map/DeckGLWrapper.tsx
satellitestudiodesign: if we add this, it needs to happen only when printing, because it makes the map look pixelated in high-density displays

GeorgiaDesigns: Oh okay, I added it to avoid redimensioning based on my display, because on the of resolution of 1920x1200, the map printed normally, but on the resolution of 1920x1080, the vertical distortion happened so I wanted to avoid using the device's display. Anyway it didn't make much differnce so I'll remove it

apps/fishing-map/features/app/App.module.css
GeorgiaDesigns: image

GeorgiaDesigns: this is so weird because on both my monitors it is keeping the aspect ratio and not warping

apps/fishing-map/features/map/DeckGLWrapper.tsx
```diff
@@ -103,7 +103,31 @@ const DeckGLWrapper = () => {
     }
     return true
   }, [])
+  const handleBeforePrint = () => {
+    const deckCanvas = document.querySelector('canvas')
```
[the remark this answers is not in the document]
GeorgiaDesigns: I just removed it because I think it didn't make much difference, I think what was distorting the map was just the styles when printing

GeorgiaDesigns:
For me, the map isn't warping at all, I tested changing my resolution and all is well, so test it out again on your monitor and let me know if it is still warping

PR #3249, opened 2025-06-17
add new layer library layers

j8seangel:
- [x] eez-areas-12nm
- [x] high-seas-pockets
- [x] paa-duke
- [] gfcm-fao


j8seangel:
merging and addressing `gfcm-fao` in another PR so translations are sooner in crowdin

PR #3391, opened 2025-09-15
duplicate vessel identity selectors fix

GeorgiaDesigns:


j8seangel (approved):
Works great, and thanks for the fix about the start and end dates!
Please take a look at the commits I pushed for review

PR #3677, opened 2026-02-16
Fishing map/router refactor

j8seangel:
[Preview branch](https://dev-ui-fishing-map-router-refactor-706952489382.us-central1.run.app/map/)

PR #3810, opened 2026-03-18
Strapi integration

GeorgiaDesigns:
[Preview link](https://dev-ui-fishing-map-strapi-integration-706952489382.us-central1.run.app/map/?sidePanelContent=userGuide&longitude=26&latitude=19&zoom=1.49)


apps/fishing-map/routes/api/corrections.ts
```diff
@@ -92,7 +92,7 @@ export const Route = createFileRoute('/api/corrections')({
           }
 
           try {
-            await sheet.loadHeaderRow(3)
+            await sheet.loadHeaderRow(4)
```
j8seangel, line 95: Has something changed in the track corrections spreadsheet?

apps/fishing-map/features/modals/Modals.tsx
```diff
@@ -89,6 +89,8 @@ const ResetWorkspaceConfig = {
   },
 }
 
+export const getModalParent = () => document.getElementById('app-layout-content') as HTMLElement
```
j8seangel, line 92: why is this needed ? and could we ensure the `app-layout-content' id is exported to a const and reused [here](https://github.com/GlobalFishingWatch/frontend/pull/3810/changes#diff-c478742f602295b41f9cc1f820159f73d746cdfd24e2f156726c167a6811c399R216)

apps/fishing-map/routes/__root.tsx
```diff
@@ -154,19 +154,6 @@ function RootComponent() {
           <Outlet />
         </I18nSSRProvider>
       </Suspense>
-      {import.meta.env.DEV && (
```
j8seangel, line 170: do you think this is not needed or useful?

libs/react-hooks/src/use-small-screen/use-small-screen.ts
```diff
@@ -1,28 +1,21 @@
...
-      window.addEventListener('resize', onWindowResize, { passive: true })
-      return () => {
-        window.removeEventListener('resize', onWindowResize)
-      }
-    }
-    return
-  }, [onWindowResize])
-
-  return isSmallScreen
+  return useSyncExternalStore(
```
j8seangel, line 16: I didn't know about this, but it's looking great 👏

apps/fishing-map/routes/_app.tsx
```diff
@@ -1,22 +1,59 @@
-import { lazy, Suspense, useEffect, useState } from 'react'
+import { Suspense, useEffect, useState } from 'react'
 import { Provider } from 'react-redux'
 import { createFileRoute, useRouter } from '@tanstack/react-router'
+
+import App from 'features/app/App'
```
j8seangel, line 6: any reason we can't load the app component with lazy loading?

j8seangel (commented):
I haven't reviewed the features/content folder yet, but just some comments about other related changes

apps/fishing-map/routes/api/corrections.ts
GeorgiaDesigns: Yes, Chris added an auto format line in the sheet that shifted the rows down

apps/fishing-map/features/modals/Modals.tsx
GeorgiaDesigns: This is needed so the modal doesn't overlay the content panel, so the user can do any action in the map without blocking the view of the user guide

apps/fishing-map/routes/__root.tsx
GeorgiaDesigns: do you think it is? I can add it back

libs/react-hooks/src/use-small-screen/use-small-screen.ts
GeorgiaDesigns: This is a fix for an issue when loading the page it would load as a small screen before the window was defined and everything would break and flash before finally hydrating. We also need to change to lazy loading from tanstack router instead of using the react one to avoid hydration mismatches

apps/fishing-map/routes/_app.tsx
GeorgiaDesigns: We should start using the [tanstack lazy loading](https://tanstack.com/router/v1/docs/api/router/lazyRouteComponentFunction) because of SSR to avoid hydration mismatches
there is also this [lazy route](https://tanstack.com/router/latest/docs/guide/code-splitting#using-the-lazytsx-suffix) option but still have to read more about it

apps/fishing-map/routes/api/corrections.ts
j8seangel: Ahh ok! is it also done in develop ?

apps/fishing-map/features/modals/Modals.tsx
j8seangel: great catch! thanks

apps/fishing-map/routes/api/corrections.ts
GeorgiaDesigns: Yes! It was done there first

apps/fishing-map/routes/_app.tsx
j8seangel: @GeorgiaDesigns I've applied the tanstack lazy loading [here](https://github.com/GlobalFishingWatch/frontend/pull/3677/changes/75ddd9f097c8c36a4a1e183d9594c0b63ee425f2) thanks !

PR #3842, opened 2026-03-26
Create integration tests github action on PRs

paciook:


j8seangel:
@paciook I've also removed the legacy run [here](https://github.com/GlobalFishingWatch/frontend/pull/3946)
thanks! this is much better 🙇

PR #3847, opened 2026-03-26
Fishing map/fix tests

GeorgiaDesigns:


j8seangel (changes requested):
However the test is still failing, this is the error I got:

```
 FAIL   fishing-map-chromium  test/integration/Polygon.spec.tsx > Polygon > should be able to draw a new polygon, see it in the map and delete it
Error: Test timed out in 30000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".

Failure screenshot:
  - apps/fishing-map/test/integration/__screenshots__/Polygon.spec.tsx/Polygon-should-be-able-to-draw-a-new-polygon--see-it-in-the-map-and-delete-it-1.png

 ❯ test/integration/Polygon.spec.tsx:98:2
     96|   })
     97|
     98|   it('should be able to draw a new polygon, see it in the map and delete it', async () => {
       |  ^
     99|     const jotaiStore = createJotaiStore()
    100|     const store = makeStore(defaultState, [], true)
```

j8seangel:
All tests passing again

image

j8seangel (approved):

PR #4001, opened 2026-05-14
Q2 usability

GeorgiaDesigns:


apps/fishing-map/features/datasets/upload/NewPointsDataset.tsx
```diff
@@ -290,6 +295,16 @@ function NewPointDataset({
           disabled={loading}
         />
       </div>
+      {isCSVFile && (startTimeProperty || endTimeProperty) && (
+        <SwitchRow
```
j8seangel, line 299: I think we should wait for @satellitestudiodesign to validate this ui change

libs/deck-layers/src/layers/user/UserPointsTileLayer.ts
```diff
@@ -213,6 +214,40 @@ export class UserPointsTileLayer<PropsT = Record<string, unknown>> extends UserB
...
 
+  private _fullBboxPromise?: Promise<Bbox | undefined>
+
+  getFullBbox = (): Promise<Bbox | undefined> => {
+    if (this._fullBboxPromise) return this._fullBboxPromise
+
+    const baseUrl = this.props.layers?.[0]?.tilesUrl
+    if (!baseUrl) return Promise.resolve(undefined)
+
+    const url = this._getTilesUrl(baseUrl).replace(/\{[zxy]\}/g, '0')
```
j8seangel: It will be more confusing for datasets filtered by time.
For example, when I click on fit bounds on a dataset that doesn't contain any points in the current timerange, it moves to an empty position. 

But also, I don't think there is a way to do it using just tiles because at zoom level 0, we might miss points because of the tile simplification... we need to rollback these changes and open a conversation with Raul to see if he has capacity to do it on the backend side

j8seangel (changes requested):
Let's wait for JC to come back to talk about the new upload workflow

PR #4018, opened 2026-05-20
User uploaded points fit bounds

GeorgiaDesigns:


j8seangel (approved):
Some tweaks for the filters and also enabled it for all layers as the API supports both points and polygons

PR #4029, opened 2026-05-25
Fishing map/sentry issues

GeorgiaDesigns:


libs/deck-layers/src/layers/fourwings/heatmap/FourwingsHeatmapTileLayer.ts
```diff
@@ -276,6 +276,7 @@ export class FourwingsHeatmapTileLayer extends CompositeLayer<FourwingsHeatmapTi
       const { comparisonMode } = this.props
       const { colorDomain: oldColorDomain } = this.state
       const newColorDomain = this._calculateColorDomain()
+      if (this.props.comparisonMode !== comparisonMode) return
```
j8seangel: I don't think this is going to happen as comparisonMode is extracted from this.props a couple of lines above

GeorgiaDesigns: It's because somewhere it seems comparisonMode is returning it's old stale value and parseing the colors using the wrong mode. I added fallbacks to handle the crash in the previous commit but the underlying issue seemed to be this

apps/fishing-map/features/workspace/workspace.slice.ts
```diff
@@ -477,6 +477,7 @@ export const updateCurrentWorkspaceThunk = createAsyncThunk<
 export function cleanReportQuery(query: QueryParams) {
   return {
     ...query,
+    bivariateDataviews: undefined,
```
j8seangel: This breaks the previous state when you go back from the report. I think this is a workaround for the stale comparisonMode but I'd like to fix without breaking changes

j8seangel (commented):
I love the idea of the diffing for the highlightedEvents dispatch 👏 I've added [this commit](https://github.com/GlobalFishingWatch/frontend/pull/4029/commits/d68a2bb81585e4a4b9cd0ae1d6475c474ce1ffbe) to avoid re generating the callback on changes to avoid re-renders and please take a look at the comment about the bivariate before merging

apps/fishing-map/features/workspace/workspace.slice.ts
GeorgiaDesigns: fixed now, I was looking at the wrong variable

PR #4123, opened 2026-06-26
Fishing map/real time proto

j8seangel:
https://fishing-map-fishing-map-real-time-proto-706952489382.us-central1.run.app/


j8seangel:
still wip but merged to develop to avoid big conflicts with the future platform changes

PR #4154, opened 2026-07-07
Composite Actions

paciook:
- **Manage Deploy vars/secrets from a sepparated JSON file**
- **Use composite actions to add a level of abstraction**


.github/workflows/build-and-push.yml
```diff
@@ -2,89 +2,34 @@ name: Build and Deploy
...
     name: Detect Affected Apps
     runs-on: ubuntu-latest
     outputs:
-      apps: ${{ steps.affected.outputs.apps }}
+      apps: ${{ steps.detect.outputs.apps }}
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
```
j8seangel, line 20: looks like there is a [version 7](https://github.com/actions/checkout/releases/tag/v7.0.0) available, why not using it?

.github/apps/production.yml
```diff
@@ -0,0 +1,78 @@
+build_env_vars: &build_env_vars
+  VITE_API_GATEWAY: https://gateway.api.globalfishingwatch.org
+  VITE_API_VERSION: v3
+  VITE_GOOGLE_MEASUREMENT_ID: G-R3PWRQW70G
+  VITE_GOOGLE_TAG_MANAGER_ID: GTM-KK5ZFST
+  VITE_USE_LOCAL_DATASETS: "false"
```
j8seangel, line 6: these env vars are only for the fishing-map app:
```
VITE_USE_LOCAL_DATASETS: "false"
VITE_USE_LOCAL_DATAVIEWS: "false"
VITE_WORKSPACE_ENV: production
VITE_REPORT_DAYS_LIMIT: "366"
VITE_PIPE_DATASET_VERSION: "4"
```
 I'd move them to its config instead of the shared one

j8seangel (commented):
👏 looks great! much cleaner and organized, thanks!

this should react to every commit in the develop branch, isn't it ? because [this commit](https://github.com/GlobalFishingWatch/frontend/commit/28a314d08ee61c39850addc0f5febfd0dabfb850) didn't trigger it yesterday

PR #4157, opened 2026-07-08
Bump checkout action + move VITE env vars to corresponding apps

paciook:


.github/actions/detect-affected/action.yml
```diff
@@ -48,7 +48,7 @@ runs:
           --base=${{ inputs.base }} \
           --head=${{ inputs.head }} \
           --type app \
-          --exclude='linting,port-labeler,*-e2e' \
```
paciook, line 51: @j8seangel I assumed this was wrong... Why wouldn't you want to trigger the port-labeler build?

This is how it was configured in `cloudbuild/template/scripts/affected-apps.sh:L14`.Let me know if I should change it back.

j8seangel:
Looks great, thanks!

PR #4169, opened 2026-07-15
redirect to settings page from map

GeorgiaDesigns:


apps/fishing-map/features/user/UserInfo.tsx
```diff
@@ -115,7 +116,10 @@ function UserInfo() {
             <p>{`${userData.firstName} ${userData.lastName || ''}`}</p>
             <p className={styles.secondary}>{userData.email}</p>
           </div>
-          <LogoutButton />
+          <div className={styles.buttonContainer}>
+            <LogoutButton />
+            <SettingsButton />
```
j8seangel, line 120: can we swap the order ? using settings first

PR #4183, opened 2026-07-21
Batch user feedback fixes

GeorgiaDesigns:


libs/deck-layers/src/layers/user/UserBaseLayer.ts
```diff
@@ -302,26 +302,12 @@ export abstract class UserBaseLayer<
 
   _getSublayerFilterExtensionProps(sublayer: ContextSubLayerConfig): FilterExtensionProps {
     if (hasSublayerFilters(sublayer) && supportDataFilterExtension(sublayer)) {
-      const filterEntries = Object.entries(getValidSublayerFilters(sublayer)) as Entries<
-        typeof sublayer.filters
-      >
-      const hasMultipleFilters = filterEntries.length > 1
```
j8seangel, line 308: I think this was needed to support filtering by timerange and custom properties, how does it work now?

j8seangel (commented):
great catch about the missing filterOperator and the thinning re-render issue

PR #4187, opened 2026-07-22
remove develop triggers

j8seangel:
As triggers now are managed by github actions, this PR removes the not used develop triggers

The terraform was already applied @paciook


paciook (approved):
LGTM

PR #4207, opened 2026-07-28
Fishing map/fix latest warning

j8seangel:


libs/datasets-client/src/datasets.utils.ts
```diff
@@ -18,6 +18,33 @@ export const getDatasetVersion = (datasetId: string) => {
...
+ */
+export const getIsDatasetVersionDowngrade = (deprecatedId: string, latestId: string) => {
+  if (!deprecatedId || !latestId) {
+    return false
+  }
+  if (removeDatasetVersion(deprecatedId) !== removeDatasetVersion(latestId)) {
+    return false
+  }
+  const from = parseDatasetVersion(deprecatedId)
+  const to = parseDatasetVersion(latestId)
```
GeorgiaDesigns:
```suggestion
  const deprecatedVersion = parseDatasetVersion(deprecatedId)
  const latestVersion = parseDatasetVersion(latestId)
```

libs/datasets-client/src/datasets.utils.ts
```diff
@@ -18,6 +18,33 @@ export const getDatasetVersion = (datasetId: string) => {
...
+  if (!from.length || !to.length || from.some(isNaN) || to.some(isNaN)) {
+    return false
+  }
+  for (let i = 0; i < Math.max(from.length, to.length); i++) {
```
GeorgiaDesigns, line 39: https://www.npmjs.com/package/semver-compare

should we use a library instead of writing this from scratch?

j8seangel: Let's keep this small piece of code and ask API team to validate directly in the deprecation headers

PR #4225, opened 2026-08-03
replace react-markdown with tanstack

j8seangel:
image
image

### DIFF
--- a/apps/platform/features/_map/content-panel/markdown-highlighter.ts
+++ b/apps/platform/features/_map/content-panel/markdown-highlighter.ts
@@ -1,0 +1,37 @@
+import { createHighlighter } from '@tanstack/highlight/core'
+import { html } from '@tanstack/highlight/languages/html'
+import { plaintext } from '@tanstack/highlight/languages/plaintext'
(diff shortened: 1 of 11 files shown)


apps/platform/features/_map/content-panel/markdown-highlighter.ts
```diff
@@ -0,0 +1,37 @@
+import { createHighlighter } from '@tanstack/highlight/core'
+import { html } from '@tanstack/highlight/languages/html'
+import { plaintext } from '@tanstack/highlight/languages/plaintext'
+import { createTanStackMarkdownHighlighter } from '@tanstack/highlight/markdown'
+import type { CodeHighlighter } from '@tanstack/markdown'
+
+// import { css } from '@tanstack/highlight/languages/css'
```
GeorgiaDesigns, line 7: What are all these commented out imports for?

j8seangel: we are not using it for now but I found useful in case we need for data-download-portal code migration for the future

PR #4265, opened 2026-08-10
try to remove not used _many translations

j8seangel:


GeorgiaDesigns (approved):
It is working well! Looks great!

PR #4267, opened 2026-08-11
Platform/help hub

GeorgiaDesigns:


apps/platform/features/help/helpHub.i18n.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { HelpHubSectionId } from './helpHub.content'
+
+export function getHelpHubSectionCopy(id: HelpHubSectionId): {
+  title: string
+  description: string
+} {
+  switch (id) {
+    case 'toolsAndFeatures':
+      return {
+        title: t((s) => s.helpHub.sections.toolsAndFeatures.title),
```
j8seangel: I think this is too nested and complicated, what about:
1. remove the `sections` key
2. make this easier to read using dynamic accessor like:
```t((s) => s.helpHub[HelpHubSectionId]title)```

so the entire function could be just:

```js
 return {
        title: t((s) => s.helpHub[id].title),
        description: t((s) => s.helpHub[id].description),
      }
```

apps/platform/features/help/helpHub.types.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Bodies are markdown but render with allowHtml, so media can be either syntax.
+// One alternation keeps whichever comes first in the body, no g flag to avoid a shared lastIndex.
+const BODY_IMAGE_REGEX =
+  /!\[([^\]]*)\]\(\s*(?:<([^>]*)>|([^\s)]+))[^)]*\)|<img\b[^>]*\bsrc=["']([^"']+)["'][^>]*>/i
+const HTML_ALT_REGEX = /\balt=["']([^"']*)["']/i
+
+// Fall back to the first image in the body when the CMS entry has no thumbnail.
+const getFirstBodyImage = (body?: string): HelpHubCard['image'] => {
```
j8seangel: please move this helpers out of the types definition and create its own file

apps/platform/features/help/helpHub.types.ts
```diff
@@ -0,0 +1,102 @@
...
+  return { url, alt: alt || undefined }
+}
+
+// Card takes { url, alt }; Strapi media carries the text as alternativeText.
+const toCardImage = (thumbnail?: StrapiImage, body?: string): HelpHubCard['image'] =>
+  thumbnail
+    ? { url: thumbnail.url, alt: thumbnail.alternativeText ?? undefined }
+    : getFirstBodyImage(body)
+
+export function toUserGuideCards(sections: UserGuideContent): HelpHubCard[] {
```
j8seangel: I think we could avoid this duplicated helpers + the custom type and convert it directly with in the component

apps/platform/features/help/HelpHubLandingPage.module.css
```diff
@@ -0,0 +1,114 @@
...
+  gap: var(--space-L);
+}
+
+.description {
+  font: var(--font-M);
+  color: var(--color-secondary-blue);
+  max-width: 70ch;
+}
+
+.seeMore {
```
j8seangel, line 37: would be great to reuse the existing ui-components secondary button. as we need to support a Tanstack Link we could extend the component but I'd like to avoid duplicating styles

apps/platform/features/help/HelpHubLandingPage.module.css
```diff
@@ -0,0 +1,114 @@
...
+  color: var(--color-secondary-blue);
+  background-color: var(--color-terthiary-blue);
+  transition: background-color 160ms ease;
+}
+
+.seeMore:hover {
+  color: var(--color-primary-blue);
+}
+
+.cardGrid {
```
j8seangel: I love this!! it is great, what about extracting to a carousel component? becuase it is going to be used in lots of places

apps/platform/features/help/HelpHubLandingPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import styles from './HelpHubLandingPage.module.css'
+
+type SectionCards = { cards: HelpHubCard[]; isLoading: boolean }
+
+function HelpHubLandingPage() {
+  const { t, i18n } = useTranslation()
+  const locale = i18n.language as Locale
+
+  const { data: userGuide = [], isLoading: isLoadingUserGuide } = useGetUserGuideQuery({ locale })
```
j8seangel: It used to work well for the map, but for the platform content we want to do SSR first so we'll need to enforce this https://tanstack.com/router/latest/docs/guide/data-loading

apps/platform/features/help/HelpHubSectionPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </div>
+    )
+  }
+
+  if (isError || !topics.length) {
+    return <p className={styles.placeholder}>{t((s) => s.common.noData)}</p>
+  }
+
+  return (
+    <div className={styles.page}>
```
j8seangel, line 57: can we extract this to a new component ?

apps/platform/features/help/HelpHubTopicContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+import ContentMarkdown from 'features/_map/content-panel/ContentMarkdown'
+import type { HelpHubTopic } from 'features/help/helpHub.types'
+
+import styles from './HelpHubTopicContent.module.css'
+
+type HelpHubTopicContentProps = {
+  topic: HelpHubTopic
+}
+
+function HelpHubTopicContent({ topic }: HelpHubTopicContentProps) {
```
j8seangel: as we'll have another section called Topic, can you rename this to HelpHubSectionContent ? or similar ?

libs/ui-components/src/card/Card.tsx
```diff
@@ -0,0 +1,37 @@
...
+export type CardTitleTag = 'h2' | 'h3' | 'h4' | 'span'
+
+export interface CardProps {
+  title: string
+  subtitle?: string
+  image?: { url: string; alt?: string }
+  titleTag?: CardTitleTag
+  className?: string
+  children?: React.ReactNode
+  testId?: string
```
j8seangel, line 11: let's support a loading state here so we can use it as placeholder while it loads like the reports placeholders where a soft animation waves

apps/platform/features/help/HelpHubLandingPage.module.css
j8seangel: I saw you already created a Carousel component that is not being used

j8seangel (changes requested):
Would be great also to research how we can improve the SSR to avoid the strapi requests as much as possible and serve already generated content

apps/platform/features/help/HelpHubLandingPage.module.css
GeorgiaDesigns: Yes I had created a Carousel, but then I realised in the designs that it doesn't have a scroll, it was supposed to be just the few cards shown, right?

GeorgiaDesigns:
Fixed most of the requested changes, but I feel we still need to discuss further the page scroll and the caroussel, so converting to draft for now so we can discuss when we are back from OOO

apps/platform/features/_map/content-panel/data-terminology/DataTerminologyContent.tsx
```diff
@@ -20,7 +20,7 @@ const DataTerminologyContent = () => {
 
   const { data, isLoading, isError } = useGetDataTerminologyContentQuery({
     id: sidePanelId as string,
-    locale: i18n.language as Locale,
+    locale: toContentLocale(i18n.language),
```
GeorgiaDesigns, line 23: Good catch

apps/platform/features/help/HelpHubSectionPage.tsx
GeorgiaDesigns: I don't think its worth extracting to a component, we only use it here and it is only for styling, I added styling to remove this div to simplify

apps/platform/features/cms/loaders/data-terminology.ts
```diff
@@ -10,12 +10,12 @@ export const getDataTerminologyContent = createServerFn({
 })
   .validator((params: { id: string; locale?: Locale; page?: number }) => params)
   .handler(({ data: { id, locale } }): Promise<StrapiResponse<DataTerminology>> => {
-    return findWithLocaleFallback<DataTerminology>(
-      'data-terminologies',
-      {
+    return fetchStrapiCollectionCached<DataTerminology>({
```
GeorgiaDesigns, line 13: I didn't quite understand what this nocache is for? Why is it needed for the preview, is it to avoid the redis cache?

j8seangel: no, this is a cache to avoid requesting the content to strapi on every user request

GeorgiaDesigns: right, makes sense

satellitestudiodesign (approved):

PR #4301, opened 2026-08-25
Platform/pipe 5 feedback

GeorgiaDesigns:


apps/platform/features/_vessels/vessel/activity/activity-by-type/ActivityByType.tsx
```diff
@@ -126,7 +126,6 @@ function ActivityByType() {
     return (
       <GroupedVirtuoso
         ref={virtuosoRef}
-        useWindowScroll
```
j8seangel, line 129: Can you share a screenshot of what you see in windows? 
or the differences between using it and not? it is weird becuause looks the same for mac

j8seangel (commented):
good improvements, thanks!

apps/platform/features/_vessels/vessel/activity/activity-by-type/ActivityByType.tsx
GeorgiaDesigns: There isn't a difference, its because it was overkill as [this one](https://virtuoso.dev/react-virtuoso/api-reference/grouped-virtuoso/#usewindowscroll-boolean) does the same as the [custom scroll](https://virtuoso.dev/react-virtuoso/api-reference/grouped-virtuoso/#customscrollparent-htmlelement)

PR #4335, opened 2026-09-03
add react compiler

j8seangel:


j8seangel:
Not doing this for now

